PR #1, opened 2020-10-29
feat[g2]: Creation of object that will be displayed on map

ogikuzma:
Creation of a parametrizable map object that will be displayed on the WPF canvas.

### DIFF
--- a/HealthcareSystem/GraphicalEditor/Enumerations/MapObjectDepartment.cs
+++ b/HealthcareSystem/GraphicalEditor/Enumerations/MapObjectDepartment.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GraphicalEditor.Enumerations
+{
+    public enum MapObjectDepartment
+    {
+        GENERAL_MEDICINE,
+        PULMOLOGY,
+        NEUROLOGY,
+        CARDIOLOGY
+    }
+}
--- a/HealthcareSystem/GraphicalEditor/Enumerations/MapObjectTypes.cs
+++ b/HealthcareSystem/GraphicalEditor/Enumerations/MapObjectTypes.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GraphicalEditor.Enumerations
+{
+    public enum MapObjectTypes
+    {
+        EXAMINATION_ROOM,
+        OPERATION_ROOM,
+        WAITING_ROOM,
+        PARKING,
+        RESTAURANT,
+        WC,
+    }
+}
--- a/HealthcareSystem/GraphicalEditor/Enumerations/MapObjectTypes.cs
+++ b/HealthcareSystem/GraphicalEditor/Enumerations/MapObjectTypes.cs
@@ -8,8 +8,10 @@ namespace GraphicalEditor.Enumerations
 {
     public enum MapObjectTypes
     {
+        BUILDING,
         EXAMINATION_ROOM,
         OPERATION_ROOM,
+        HOSPITALIZATION_ROOM,
         WAITING_ROOM,
         PARKING,
         RESTAURANT,
--- a/HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/Building.cs
+++ b/HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/Building.cs
@@ -1,0 +1,22 @@
+﻿using GraphicalEditor.Enumerations;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GraphicalEditor.Models.MapObjectRelated
+{
+    class Building : MapObjectEntity
+    {
+        private int _numOfFloors;
+
+        public Building(String description, int numOfFloors)
+            : base(new MapObjectType(MapObjectTypes.BUILDING), "Building 1")
+        {
+            this.NumOfFloors = numOfFloors;
+        }
+
+        public int NumOfFloors { get => _numOfFloors; set => _numOfFloors = value; }
+    }
+}
--- a/HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectType.cs
+++ b/HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectType.cs
@@ -1,0 +1,39 @@
+﻿using GraphicalEditor.Enumerations;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Media;
+
+namespace GraphicalEditor.Models.MapObjectRelated
+{
+    public class MapObjectType
+    {
+        private MapObjectTypes _mapObjectType;
+
+        public MapObjectType(MapObjectTypes mapObjectType)
+        {
+            this._mapObjectType = mapObjectType;
+        }
+
+        public SolidColorBrush getColor()
+        {
+            switch (this._mapObjectType)
+            {
+                case MapObjectTypes.EXAMINATION_ROOM:
+                    return Brushes.Green;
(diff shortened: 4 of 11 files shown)


HealthcareSystem/GraphicalEditor/Enumerations/MapObjectDepartment.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace GraphicalEditor.Enumerations
+{
+    public enum MapObjectDepartment
+    {
+        GENERAL_MEDICINE,
+        PULMOLOGY,
+        NEUROLOGY,
+        CARDIOLOGY
+    }
```
milana-todorovic, line 15: Consider using a class instead of enum for departments, the current solution isn't very flexible. Possibly refactor this at a later point when a connection with data storage is established, I assume that the current priority is just making objects that can be drawn on a map?

HealthcareSystem/GraphicalEditor/Enumerations/MapObjectTypes.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public enum MapObjectTypes
+    {
+        EXAMINATION_ROOM,
+        OPERATION_ROOM,
+        WAITING_ROOM,
+        PARKING,
+        RESTAURANT,
+        WC,
+    }
```
milana-todorovic, line 19: Rooms for hospitalization should be added.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectType.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Media;
+
+namespace GraphicalEditor.Models.MapObjectRelated
+{
+    public class MapObjectType
+    {
```
milana-todorovic, line 12: Putting colors in the type seems somewhat odd, since it looks like the MapObjectBasics class was generally intended to contain properties of the room that aren't strictly related to drawing? I would consider defining a ColorProvider interface that takes the type as an argument and returns a color, although using that without a DI container might be inconvenient. Definitely consider moving the color information elsewhere eventually though.

milana-todorovic (changes requested):
Changes are requested because hospitalization rooms are missing from the MabObjectTypes enum, the comments with refactoring ideas are mostly for later consideration and discussion.

HealthcareSystem/GraphicalEditor/Enumerations/MapObjectDepartment.cs
ogikuzma: I can convert it to a class. I intended to do it later because there could be one reason why it is more convenient to use it as a class for now, but I can do it now if you insist.
Making an object that could be drawn on a map was part of the proof of concept which has already been done. Now, we are working on a more flexible map object.

HealthcareSystem/GraphicalEditor/Enumerations/MapObjectTypes.cs
ogikuzma: Ok, I'll do it. Our product owner hasn't mentioned hospitalization rooms when she was explaining that part, but I guess she has planned to leave for the next sprint.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectType.cs
ogikuzma: In the meantime, MapObjectBasics has been converted into another class and there has been a generalization introduced, so there are some differences from the current state. I'll leave the colors here for now. I believe that this works quite as intended for now and probably for the future. If necessary, refactoring will be done in the future.

ogikuzma (commented):
Answers to the comments.

HealthcareSystem/GraphicalEditor/Enumerations/MapObjectTypes.cs
milana-todorovic: Sure, I assumed it was an accidental omission since all the other room types that exist on the backend are added already. I can't really see any harm in adding it right now though, and later it might be harder to do if more places that use it like the switch for colors appear?

ogikuzma: Yes, that's right. It will surely be needed for medicine and equipment manipulation later in the future.

HealthcareSystem/GraphicalEditor/Enumerations/MapObjectDepartment.cs
milana-todorovic: Whenever works for you really, I added comments about refactoring just for later consideration since the current code is mostly intended as proof of concept. (I probably shouldn't have put them in a review that requests changes though, sorry about that. I'll do it separately in the future.)

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectType.cs
```diff
@@ -0,0 +1,43 @@
...
+        private MapObjectTypes _mapObjectType;
+
+        public MapObjectType(MapObjectTypes mapObjectType)
+        {
+            this._mapObjectType = mapObjectType;
+        }
```
vlaksi, line 18: ### Comment

For more clear code in C# you should avoid use **this** identifier

image

### Proposal 

Remove all this identifiers from code

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/Building.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace GraphicalEditor.Models.MapObjectRelated
+{
+    class Building : MapObjectEntity
+    {
+        private int _numOfFloors;
+
+        public Building(String description, int numOfFloors)
+            : base(new MapObjectType(MapObjectTypes.BUILDING), "Building 1")
+        {
+            this.NumOfFloors = numOfFloors;
```
milana-todorovic, line 17: Using 'this' here isn't necessary.

milana-todorovic (approved):
Looks good to me for now, I like the idea of adding a generalization here.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectType.cs
ogikuzma: Although using **this** is pretty much redundant in these cases, I really like using it since it helps me to quickly distinguish class fields from other variables, but I will avoid using it in this project if you don't like that behavior.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/Building.cs
ogikuzma: Although using **this** is pretty much redundant in these cases, I really like using it since it helps me to quickly distinguish class fields from other variables, but I will avoid using it in this project if you don't like that behavior.

vlaksi (approved):
Lgtm, only remove usage of 'this' , and we are good for now

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectType.cs
vlaksi: If you are using ' _ ' it say that is local, so i think it is enough to use '_' or 'this' (if you prefer it). But both is not good practice.

milana-todorovic: I don't really mind **this** in constructors, but we should definitely decide on a single practice for everyone and do it consistently. Maybe put up a vote?

vlaksi: A single practice for everyone is a good idea. We should up a vote at the next meeting.

ogikuzma: I agree. Good suggestion.

VasilijeBursac:
[(2) g2: Create & display basic objects](https://trello.com/c/61L8sQPg/5-2-g2-create-display-basic-objects)

PR #4, opened 2020-10-29
feat[g3]: Creation of map objects door

ogikuzma:
Creation of map objects door and making examples of MapObject use.

### DIFF
--- a/HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
+++ b/HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
@@ -31,29 +31,13 @@ namespace GraphicalEditor.Models.MapObjectRelated
         {
             Rectangle = new Rectangle();
             Rectangle.Fill = Brushes.DarkGreen;
-            Rectangle.Width = GetDoorWidth();
-            Rectangle.Height = GetDoorHeight();
+            Rectangle.Width = DoorWidth;
+            Rectangle.Height = DoorHeight;
 
             Rectangle.SetValue(Canvas.LeftProperty, CalculateDoorX());
             Rectangle.SetValue(Canvas.TopProperty, CalculateDoorY());
 
             return Rectangle;
-        }
-
-        // Rotacija vrata - zameni sirinu i visinu ukoliko vrata trebaju da se postave vertikalno ( horizontalno su po default)
-        private double GetDoorWidth()
-        {
-            if (MapObjectDoorOrientation == MapObjectDoorOrientation.TOP || MapObjectDoorOrientation == MapObjectDoorOrientation.BOTTOM)
-                return AllConstants.DOOR_WIDTH;
-            else return AllConstants.DOOR_HEIGHT;
-        }
-
-        // Rotacija vrata - zameni sirinu i visinu ukoliko vrata trebaju da se postave vertikalno (horizontalno su po default)
-        private double GetDoorHeight()
-        {
-            if (MapObjectDoorOrientation == MapObjectDoorOrientation.TOP || MapObjectDoorOrientation == MapObjectDoorOrientation.BOTTOM)
-                return AllConstants.DOOR_HEIGHT;
-            else return AllConstants.DOOR_WIDTH;
         }
 
         private double CalculateDoorX()
@@ -70,7 +54,7 @@ namespace GraphicalEditor.Models.MapObjectRelated
         }
 
         private double CalculateXForLeft()
-            => MapObjectMetrics.X - GetDoorWidth() + AllConstants.RECTANGLE_STROKE_THICKNESS;
+            => MapObjectMetrics.X - DoorWidth + AllConstants.RECTANGLE_STROKE_THICKNESS;
 
         private double CalculateXForRight()
             => MapObjectMetrics.X + MapObjectMetrics.Width - AllConstants.RECTANGLE_STROKE_THICKNESS;
@@ -89,7 +73,7 @@ namespace GraphicalEditor.Models.MapObjectRelated
         }        
 
         private double CalculateYForTop()
-            => MapObjectMetrics.Y - GetDoorHeight() + AllConstants.RECTANGLE_STROKE_THICKNESS;
+            => MapObjectMetrics.Y - DoorHeight + AllConstants.RECTANGLE_STROKE_THICKNESS;
 
         private double CalculateYForBottom()
             => MapObjectMetrics.Y + MapObjectMetrics.Height - AllConstants.RECTANGLE_STROKE_THICKNESS;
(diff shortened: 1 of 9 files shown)


HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        private Rectangle _rectangle;
+        private MapObjectMetrics _mapObjectMetrics;
+        private MapObjectDoorOrientation _mapObjectDoorOrientation;
+        private double _XShift;
+        private double _YShift;
+
+        private readonly double DOOR_WIDTH = 45;
+        private readonly double DOOR_HEIGHT = 10;
+
```
vlaksi: Could you make one file with all constants? For an easy change in the future.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
```diff
@@ -0,0 +1,114 @@
...
+namespace GraphicalEditor.Models.MapObjectRelated
+{
+    public class MapObjectDoor
+    {
+        private Rectangle _rectangle;
+        private MapObjectMetrics _mapObjectMetrics;
+        private MapObjectDoorOrientation _mapObjectDoorOrientation;
+        private double _XShift;
+        private double _YShift;
+
```
vlaksi: What are these fields used for? Maybe give a more descriptive name?

ogikuzma: By default, a door is positioned in the center of the border/wall defined by MapObjectDoorOrientation (top, bottom, left, right). If you would like to shift the door by X or Y, you could use those variables to do so.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
ogikuzma: Ok, no problem.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
ogikuzma: I cannot think of any better name. I could add a comment above them since that is the best way to understand the real usage.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
```diff
@@ -0,0 +1,114 @@
...
+            switch (MapObjectDoorOrientation)
+            {
+                case MapObjectDoorOrientation.TOP:
+                    return MapObjectMetrics.Y - GetDoorHeight() + MapObject.RECTANGLE_STROKE_THICKNESS;
+                case MapObjectDoorOrientation.BOTTOM:
+                    return MapObjectMetrics.Y + MapObjectMetrics.Height - MapObject.RECTANGLE_STROKE_THICKNESS;
+                default:
+                    return CalculateYShifted();
+            }
+        }
```
vlaksi: ### Comment

Default and other cases are not at the same level of abstraction.

### Proposal

```sh
        private double CalculateDoorY()
        {
            switch (MapObjectDoorOrientation)
            {
                case MapObjectDoorOrientation.TOP:
                    return CalculateDoorTopForY();
                case MapObjectDoorOrientation.BOTTOM:
                    return CalculateDoorBottomForY();
                default:
                    return CalculateYShifted();
            }
        }

       private double CalculateDoorTopForY(){
           return MapObjectMetrics.Y - GetDoorHeight() + MapObject.RECTANGLE_STROKE_THICKNESS;
       }

       private double CalculateDoorBottomForY(){
           return MapObjectMetrics.Y + MapObjectMetrics.Height - MapObject.RECTANGLE_STROKE_THICKNESS;
       }

```

I give a proposal for Y but, it is the same proposal for X.

vlaksi (changes requested):
You will find the comments inline, including the proposal.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
ogikuzma: I agree, it will be refactored.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
vlaksi: ### Comment

Maybe like this:

```sh
        private double _XShiftFromCenter;
        private double _YShiftFromCenter;
```

### Proposal

Try to avoid comment, better is to put a long descriptive name.

vlaksi (dismissed):
Lgtm

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
```diff
@@ -0,0 +1,114 @@
...
+        private readonly double DOOR_HEIGHT = 10;
+
+        public MapObjectDoor(MapObjectDoorOrientation mapObjectDoorOrientations, double XShift = 0, double YShift = 0)
+        {
+            MapObjectDoorOrientation = mapObjectDoorOrientations;
+            this.XShift = XShift;
+            this.YShift = YShift;
+        }
+
+        public Rectangle GetDoor()
```
milana-todorovic, line 30: Change proposal: Replace the GetDoor, GetDoorWidth, and GetDoorHeight methods with properties. I believe they don't violate any of the usage guidelines outlined in the [Properties vs. Methods section here](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/bzwdh01d(v=vs.71)?redirectedfrom=MSDN#cpconpropertyusageguidelinesanchor1), especially GetDoorWidth and GetDoorHeight. GetDoor might requite some further consideration since it is a somewhat more expensive operation.

milana-todorovic (changes requested):

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectDoor.cs
ogikuzma: I'll convert GetDoorWidth and GetDoorHeight to properties since it is a good suggestion but will leave GetDoor since I don't like long properties especially if they rely on other methods.

milana-todorovic: Sounds good, I agree that GetDoor is better left as a method.

milana-todorovic (approved):

vlaksi (approved):
lgtm

PR #6, opened 2020-10-30
feat[g2]: Created id generator for map objects

djura023:
Added IdGenerator class,which automatically generates an id for the map object using objectsList count.Objects list is added to Main Window for now,because we dont have an implemented data base.


anciz (approved):
Everything seems to be OK for now.

vlaksi:
@djura023 

You could just add DevOps for code review, and someone will review it. You do not need explicitly to choose which member will do it.  

image

vlaksi (approved):
lgtm

PR #7, opened 2020-10-30
feat[g1]: Access hospital map from current WPFs

djura023:


VasilijeBursac:
[(1) g1: Access hospital map from current WPFs](https://trello.com/c/s3SIyZKO/13-1-g1-access-hospital-map-from-current-wpfs)

HealthcareSystem/ManagerFrontend/MainWindow.xaml.cs
```diff
@@ -208,6 +208,12 @@ private void Navigation_Drugs_Click(object sender, RoutedEventArgs e)
...
 
-		#endregion
-	}
+        #endregion
+
+        private void Navigation_Graphical_Editor_Click(object sender, RoutedEventArgs e)
+        {
+			GraphicalEditor.MainWindow mw = new GraphicalEditor.MainWindow();
+			mw.ShowDialog();
+        }
```
vlaksi, line 217: LGTM, but if you only use one MW, try to remove it for a more clean function.

```sh
        private void Navigation_Graphical_Editor_Click(object sender, RoutedEventArgs e)
        {
		GraphicalEditor.MainWindow().ShowDialog();;
        }
```

vlaksi: But if you want, you can keep it, but give it a more descriptive name.

```sh
        private void Navigation_Graphical_Editor_Click(object sender, RoutedEventArgs e)
        {
		GraphicalEditor.MainWindow mainWindow = new GraphicalEditor.MainWindow();
		mainWindow.ShowDialog();
        }
```

vlaksi (changes requested):
In general, try to give more descriptive names to variables, functions, classes. It is okay if variable, function, class name have 3 and more words. With that, we are getting a more readable code.

What is 'mp2' used for?

image

What is 'objectsList' used for? Also, avoid using the type of data in the names of variables. Because when we change the type of variable structure, then your code does not tell the truth anymore.

image

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/IdGenerator.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GraphicalEditor.Models.MapObjectRelated
+{
+   public  class IdGenerator
+    {
+        private long _id;
```
milana-todorovic: What is the _id field in IdGenerator currently used for?

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/IdGenerator.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GraphicalEditor.Models.MapObjectRelated
+{
+   public  class IdGenerator
```
milana-todorovic: What is the responsibility of this class?

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/IdGenerator.cs
```diff
@@ -0,0 +1,30 @@
...
+        private long _id;
+
+        public IdGenerator()
+        {
+            _id = GetNextID();
+        }
+
+        public long GetNextID()
+        {
+            MainWindow wnd = (MainWindow)App.Current.MainWindow;
```
milana-todorovic: Why does the IdGenerator class need to access the window?

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectEntity.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace GraphicalEditor.Models.MapObjectRelated
 {
-    public class MapObjectEntity
+    public class MapObjectEntity : IdGenerator
```
milana-todorovic: Why does MapObjectEntity inherit from IdGenerator?

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/IdGenerator.cs
```diff
@@ -12,12 +12,12 @@ public  class IdGenerator
...
         public IdGenerator()
         {
-            _id = GetNextID();
+            _id = 0;
         }
 
         public long GetNextID()
         {
-            MainWindow wnd = (MainWindow)App.Current.MainWindow;
+            MainWindow wnd = new MainWindow();
```
milana-todorovic: I assume that this is a patch that happened because when you link to the other WPF project, the main window is no longer the one from the graphical editor. Creating a new instance means that this will not count the objects correctly at all, and you will get the same id every time. The MainWindow dependency should be removed.

milana-todorovic (changes requested):
The responsibility of the IdGenerator is unclear, and having it access the ObjectList property of the MainWindow couples them and creates an odd dependency. I assume that the purpose was just to create unique ids for the map objects. I think that a better way to do that at this point would be to just add an id field to the map object class that needs them, and then create a simple id generator that works as a counter. Then you could keep an instance of the generator in the MainWindow class that currently creates the map objects, and whenever a new object is created, generate an id and set it.

vlaksi:
### Problem

If you get a description, you should use it instead of the fixed "Building 1" string.

image

### Example

image

ogikuzma:
@vlaksi I cannot answer as a reply for some reason, so I will answer this way. I've written MainWindow.xaml.cs only for presentational/educational purposes to show how objects will be created. This file will be completely rewritten after we start designing, so it should not worry you. As your last comment is concerned, "Building 1" is only written for test purposes. It will be changed with description parameter really soon.

ogikuzma:
I've made some changes to IdGenerator class. It is now called Entity. This is the approach our professor used in the last semester that we also used in our previous project and worked perfectly fine. Entity class is now coupled with MainWindow for test purposes, but as soon as we establish DB connection it will be changed. I believe that moving id generation logic to MapObjectEntity would also make sense and this is the approach we will probably consider. I hope this solves any confusion you might have had.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/Entity.cs
```diff
@@ -0,0 +1,29 @@
...
+        public long Id { get; set; }
+
+        public Entity()
+        {
+            Id = GenerateId();
+        }
+
+        private long GenerateId()
+        {
+            MainWindow mainWindow = (MainWindow)App.Current.MainWindow;
```
milana-todorovic: The main issue with the IgGenerator was this logic, going back to it is not acceptable. When you run the graphical editor from the other WPF apps you won't be able to access the window you want, because it will no longer be the main window of the app. Try starting the editor from the manager GUI and see what happens.

milana-todorovic (changes requested):
Fix the Entity class, and please merge develop to this branch to solve conflicts.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/Entity.cs
ogikuzma: I see the problem. It should be solved by this fix until we establish a connection with DB.

ogikuzma: I'll resolve conflicts.

ogikuzma:
Conflicts should be solved now. Sorry for making changes to the door during the merge. It should have been a separate commit.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -25,48 +25,53 @@ namespace GraphicalEditor
...
+                   new MapObjectDoor(MapObjectDoorOrientation.NONE)
+            );
+
             parking.AddToCanvas(_canvas);
 
-            // Dodavanje vise objekata od jednom
             for (int i = 0; i <= 10; i++)
             {
                 double startX = 20;
                 double width = 70;
```
Pufke: Everything seems fine to me, 
just one small suggestion for better understanging of code,
"width", "startX", "x"  are not good names - it is already inferred from the type. 
image

Much better approach is widthOf...name of something...
e.g: widthOfCanvas, widthOfExaminationRoom

HealthcareSystem/ManagerFrontend/MainWindow.xaml.cs
```diff
@@ -208,6 +208,12 @@ private void Navigation_Drugs_Click(object sender, RoutedEventArgs e)
 			MainFrame.Content = new DrugsView();
 		}
 
-		#endregion
-	}
+        #endregion
+
+        private void Navigation_Graphical_Editor_Click(object sender, RoutedEventArgs e)
+        {
+			GraphicalEditor.MainWindow mw = new GraphicalEditor.MainWindow();
```
Pufke: Try to come up with a more meaningful name than "mw".
Much better aproach for "mw" is graphicalEditorMainWindow.
Maybe in this block of code it is not necessary and it is clear from the context of what mw is, but our goal is to minimize the need to understand the context while reading the code, currently this is not a problem but such little things are a potential big problem in the future

Pufke: I see you did a good job in
image

Try to stay consistent and apply the same approach here

Pufke (dismissed):

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MapObjectMetrics.cs
```diff
@@ -8,22 +8,18 @@ namespace GraphicalEditor.Models.MapObjectRelated
...
         {
-            this.X = x;
-            this.Y = y;
-            this.Width = width;
-            this.Height = height;
+            XOfCanvas = xOfCanvas;
+            YOfCanvas = yOfCanvas;
+            WidthOfMapObject = widthOfMapObject;
+            HeightOfMapObject = heightOfMapObject;
         }
```
Pufke, line 22: These are much better names that completely conveys meanings.
I cannot overstate how easier it is to understand this code. 
Keep it up :)

milana-todorovic (approved):
Everything seems ok now.

vlaksi (approved):
lgtm

PR #9, opened 2020-10-31
Refactor w1.5 refactoring backend code

sladjana-savkovic:


sladjana-savkovic:

HealthcareSystem/Backend/Controller/NotificationSurveyAndFeedback/FeedbackController.cs
```diff
@@ -15,16 +15,23 @@ public class FeedbackController
...
-      }
-      
-      public List<Model.Users.Feedback> ViewFeedbackInformations()
-      {
-            return feedbackService.ViewFeedbackInformations();
-      }
-   
-   
-   }
+        public void NewFeedback(Feedback feedback)
```
milana-todorovic: Is there a specific reason to name this method 'NewFeedback' instead of 'AddFeedback'? As far as I am aware most of the existing code uses 'AddX' as the name, and since this is definitely something that should be consistent, I feel like eventually switching to 'NewX' everywhere is just unnecessary work.

HealthcareSystem/Backend/Model/Users/Feedback.cs
```diff
@@ -5,19 +5,28 @@
...
 namespace Model.Users
 {
-   public class Feedback
-   {
-        public Stars Grade { get; set; }
+    public class Feedback
+    {
+        public int Id { get; set; }
         public string Comment { get; set; }
+        public bool IsAnonymous { get; set; }
```
milana-todorovic: Is this field necessary? I feel like having a null Commentator property should be enough to make a Feedback anonymous. 
I would consider keeping the IsAnonymous property for convenience, but only the getter and calculate it based on Commentator.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
```diff
@@ -13,15 +14,22 @@ namespace Service.NotificationSurveyAndFeedback
...
-            return feedbackRepository.NewFeedback(feedback);
-      }
-      
-      public List<Model.Users.Feedback> ViewFeedbackInformations()
-      {
-            return feedbackRepository.GetAllFeedbacks();
-      }
-   
-   }
+        public void NewFeedback(Feedback feedback)
```
milana-todorovic: I feel like some sort of validation is necessary here, for example that feedback is not null, that the commentator id is valid, and so on.

HealthcareSystem/DoctorFrontend/AddFeedback.xaml.cs
```diff
@@ -28,35 +28,6 @@ public AddFeedback()
 
         private void Button_Click(object sender, RoutedEventArgs e)
         {
```
milana-todorovic: Since the Commentator property in Feedback can only be a Patient, I don't really see how it would be possible for other users to leave feedback at all at this point. In light of this I would remove the option to leave feedback from the doctor, secretary, and manager user interfaces for now, rather than leave the interface unchanged and make the send button do nothing.

milana-todorovic (changes requested):
Along with the comments I left for specific sections of code, I feel like the fact that we plan to migrate to a database soon wasn't given enough consideration while refactoring. Moving towards using the Repository Pattern would make the migration much easier, so I feel like this would be a good moment to start working on that by extracting an interface from the FeedbackRepository class.

HealthcareSystem/Backend/Repository/FeedbackRepository.cs
```diff
@@ -21,18 +21,55 @@ public FeedbackRepository()
...
             path = Path.GetFullPath(fileName);
         }
-        public Model.Users.Feedback NewFeedback(Model.Users.Feedback feedback)
+        public void NewFeedback(Feedback feedback)
         {
             List<Feedback> feedbacks = ReadFromFile();
             feedbacks.Add(feedback);
             WriteInFile(feedbacks);
-            return feedback;
         }
```
vlaksi: # Comment

You are working on refactoring backend code, and this is a great moment to apply SOLID. Think about it for a second. What is a problem with methods like this? What will happen when we need to go from FileRepository implementation to NoSQL DB? What would happen if we go to MySQL DB? It is very hard to change every function which has FileRepo to NoSQL repo implementation.

To resolve this, I suggest working on the DIP(Dependency inversion principle). Not just for this FeedbackRepository but for the whole Repository part (for the beginning working on FeedbackRepository is a good start point). Also, look at [REPOSITORY/FACADE](https://www.youtube.com/watch?v=Vb3xx0U21hk) pattern and after you apply this advice, we would have more maintainable and reliability software.

# Concrete example

If we want to change the DataBase from one to another, with this design that is easy. With dependency injection, we can inject in FeedbackServis constructor which Repository implementation we want. And if we want to move from one DataBase to another, we have to inject that another database.

## Design

image

</br> </br>

## FileRepository implementation for example

image

image

</br>

## Move to MySQL Repository implementation

image

If you see, I only change one line to switch database implementation. Isn't that beautiful.

vlaksi (changes requested):
The inline comments give concrete improvements.

HealthcareSystem/Backend/Controller/NotificationSurveyAndFeedback/FeedbackController.cs
```diff
@@ -14,22 +15,22 @@ namespace Controller.NotificationSurveyAndFeedback
    public class FeedbackController
    {
 
-        private FeedbackService feedbackService = new FeedbackService();
-        public void NewFeedback(Feedback feedback)
+        private FeedbackService _feedbackService = new FeedbackService(new MySqlFeedbackRepository());
```
milana-todorovic: I would keep the service connected to the file repository for now instead, since there is no database yet, and you'll need to access a repository that's actually implemented.

HealthcareSystem/Backend/Repository/FileFeedbackRepository.cs
```diff
@@ -1,27 +1,20 @@
...
-
-using Model.Users;
+﻿using Model.Users;
 using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.IO;
 
-namespace Repository
+namespace Backend.Repository
```
milana-todorovic: I would change the namespace name back to just Repository.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
```diff
@@ -13,22 +14,33 @@ namespace Service.NotificationSurveyAndFeedback
 {
    public class FeedbackService
    {
-        private FeedbackRepository feedbackRepository = new FeedbackRepository();
-        public void NewFeedback(Feedback feedback)
+        private IFeedbackRepository _feedbackRepository;
+
+        private ActivePatientRepository _activePatientRepository = new ActivePatientRepository();
```
milana-todorovic: Since you're using the ActivePatientRepository, maybe make an interface there as well now? That way you won't have to touch this class again when that part of the system is refactored. Also, just a style suggestion, but I would remove the empty line between the two fields.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
```diff
@@ -13,22 +14,33 @@ namespace Service.NotificationSurveyAndFeedback
...
+
+        public FeedbackService(IFeedbackRepository feedbackRepository)
+        {
+            _feedbackRepository = feedbackRepository;
+        }
+        public void AddFeedback(Feedback feedback)
         {
-            feedbackRepository.NewFeedback(feedback);
+            if (feedback == null)
+                return;
```
milana-todorovic: I would throw exceptions here instead, this way the caller doesn't know if the operation was successful or not. Maybe now would be a good moment to add custom exceptions?

milana-todorovic (changes requested):

HealthcareSystem/Backend/Service/UsersAndWorkingTime/PatientService.cs
```diff
@@ -14,16 +14,21 @@ namespace Service.UsersAndWorkingTime
...
+        }
         public Patient RegisterPatient(Patient patient)
         {
             if (!patient.IsGuest &&  (!IsUsernameValid(patient.Username) || !IsPasswordValid(patient.Password)))
             {
                 return null;
             }
-            return activePatientRepository.NewPatient(patient);
+            return _activePatientRepository.AddPatient(patient);
         }
```
vlaksi: I would also throw exceptions here instead of return null.

vlaksi (changes requested):
## Comment

Now is also a good moment for refactoring the structure Repository directory. When we open the Repository directory, it is hard to find what you want. 

image

## Proposal

Try to move similar classes into one directory. Something like this ( but you can do it how you think is the best ).

image

image

image

milana-todorovic (dismissed):

vlaksi (changes requested):
@sladjana-savkovic Could you also separate this into one directory?

image

sladjana-savkovic:
> Could you also separate this into one directory?

I could, but I am not sure if is right to have separate repository for feedback. Maybe it could merge later with SurveyRepository, such as the Feedback and survey repository.

vlaksi (approved):
You're right, we can merge Survey & Feedback into one directory.

milana-todorovic (approved):

sladjana-savkovic:
> You're right, we can merge Survey & Feedback into one directory.

Ok, I will merge it when repository pattern for survey be done. For now, feedback has own repository.

PR #11, opened 2020-11-03
feat[g4]: Building a map from basic objects

VasilijeBursac:
Please keep in mind that code in LoadMockupObjects() function is only temporary and it will be removed after we establish loading objects from file, so the code isn't the cleanest....


VasilijeBursac:
[(5) g4: Build a map from basic objects](https://trello.com/c/BiuN58fv/21-5-g4-build-a-map-from-basic-objects)

vlaksi (approved):
lgtm

milana-todorovic (approved):
Seems alright for now.

PR #13, opened 2020-11-03
refactor[w1.5]updating model and migrate to database

sladjana-savkovic:


sladjana-savkovic:

HealthcareSystem/Backend/Model/Users/City.cs
```diff
@@ -5,13 +5,17 @@
...
 
 namespace Model.Users
 {
     public class City
     {
-        public string Name { get; set; }
+        [Key]
         public int ZipCode { get; set; }
+        public string Name { get; set; }
+        public virtual ICollection<User> Users { get; set; }
```
milana-todorovic, line 18: Have you tried using the [manual configuration for one to many relationships described here](https://docs.microsoft.com/en-us/ef/core/modeling/relationships?tabs=fluent-api%2Cfluent-api-simple-key%2Csimple-key#manual-configuration) instead of adding the collection properties? I haven't tried but apparently it should be possible to use this and just leave the 'many' side empty, if it works it might be easier than doing it like this for the entire model. 
I believe it should go something like:
<code>modelBuilder.Entity\<User\>()
            .HasOne(p => p.City)
            .WithMany();</code>
I'm not sure if user being abstract will make issues here, but it should work for Feedback and Patient at least? It's alright like it is right now, but this might be an option worth considering if doing it the current way becomes too tedious when the rest of the model is migrated.

milana-todorovic (dismissed):
Everything seems alright.

anciz (dismissed):
This seems to be good enough.

sladjana-savkovic:
I implemened methods in mysql repositories. All methods in these repositories follow the rules about clean code (add or update return void etc.). I apply clean code to all feedback's layers, too.  But some classes causes problems with clean code in layers such as service and controller. Also, controllers for doctor, patient, secretary, manager are implemented using strategy pattern which is not easy to correct (code behind). The conclusion is, there is too much to refactor all methods so as to be consistent with clean code rules. I am aware of the bad things of this and suggest that all the refactoring work be done in the next few sprints. For now, all new classes or classes we need for this sprint, are clean.

milana-todorovic (commented):
Everything seems ok, just merge develop to resolve conflicts first.

anciz (dismissed):
Well done!

anciz (approved):

milana-todorovic (approved):

PR #14, opened 2020-11-03
feat[w1.1]: create frontend for add feedback

draganacarapic:
First commit contains libraries for jQuery and Bootstrap MaterialDesign on path : /wwwroot/lib.


draganacarapic:

anciz (approved):
Well done!
We only need to  decide  which language to use for interface (English or Serbian).

milana-todorovic (approved):
The UI looks very nice.

PR #19, opened 2020-11-04
feat[w1.3]: create html file for unpublished feedback

jelenazeljko55:


jelenazeljko55:
[(3) W1.3 : As a system administrator I want to publish feedback so that I could share positive experiences with other users.](https://trello.com/c/CaAvA5FG/3-3-w13-as-a-system-administrator-i-want-to-publish-feedback-so-that-i-could-share-positive-experiences-with-other-users)

HealthcareSystem/PatientWebApp/wwwroot/css/style_for_unpublished_feedbacks.css
```diff
@@ -0,0 +1,131 @@
+﻿body#add-apart, body#guests-review, body#comments {
```
milana-todorovic, line 1: As far as I can see you have Bootstrap included in the project, I feel like using that would be much easier than writing custom CSS?

milana-todorovic (approved):
The navbar isn't consistent across pages and I wasn't able to navigate to this page from the homepage, but I assume that's just because not all the pages have been added yet? Other than that everything's ok.

anciz (approved):
Seems to be OK.

PR #22, opened 2020-11-05
feat[w1.1] send ajax for create new feedback object

draganacarapic:


draganacarapic:

HealthcareSystem/PatientWebApp/wwwroot/js/add_feedback.js
```diff
@@ -0,0 +1,85 @@
...
+			$('#ap_error_msg').attr("hidden", false);
+
+			setTimeout(function () {
+				$('#ap_error_msg').attr("hidden", true);
+			}, 2000);
+
+			return;
+		}
+
+		//I take the first patient from the database, otherwise the currently logged in patient will be taken
```
milana-todorovic, line 24: This makes sense since you can't log in a patient yet, but I'm not sure it's the most painless way to do it? Just taking the first patient on the backend instead might've been easier, that way you wouldn't have to expose an endpoint for getting a patient right now. Existing code will have to be changed later either way though, so whichever suits you more is good enough.

milana-todorovic (approved):

anciz (approved):

PR #23, opened 2020-11-05
feat[w1.3]: implement backend for publishing feedback

draganacarapic:


draganacarapic:
[(3) W1.3 : As a system administrator I want to publish feedback so that I could share positive experiences with other users.](https://trello.com/c/CaAvA5FG/3-3-w13-as-a-system-administrator-i-want-to-publish-feedback-so-that-i-could-share-positive-experiences-with-other-users)

HealthcareSystem/PatientWebApp/Controllers/FeedbackController.cs
```diff
@@ -51,6 +51,21 @@ public ActionResult getPublishedFeedbacks()
             }
         }
 
+        [HttpPut]
+        public ActionResult PublishFeedback(FeedbackDTO feedbackDTO)
```
milana-todorovic: This isn't really the path I'd expect this to be at when it comes to [RESTful resource naming](https://restfulapi.net/resource-naming/), put at this path would imply updating the entire list, not just one entity. I would expect something like put at /feedback/{id}, or even put or post at /feedback/{id}/publish since it's not supposed to be a general update.

HealthcareSystem/PatientWebApp/Controllers/FeedbackController.cs
```diff
@@ -51,6 +51,21 @@ public ActionResult getPublishedFeedbacks()
...
+        {
+            try
+            {
+                Feedback feedback = FeedbackAdapter.FeedbackDTOToFeedback(feedbackDTO);
+                _feedbackService.PublishFeedback(feedback);
```
milana-todorovic: Another thing that I just realized is that the way this method is implemented in the FeedbackService isn't actually very good - it can currently be used to perform any update, not specifically publish. I feel like this branch might be a good place to fix that.

milana-todorovic (changes requested):

milana-todorovic (approved):
That's much better, everything seems ok to me now.

anciz (approved):
Alright!

PR #24, opened 2020-11-05
feat[g2]: Auto generating object description, new object colors

VasilijeBursac:
By creating function "FormatObjectDescription()" and overriding it, I tried to create Runtime Polymorphism. Probably could have used toString() instead of it but with a little bit of additional code, so it is basically comes to the same thing... I haven't used overriding of toString() method because it's name isn't really descriptive in this case...


VasilijeBursac:
[(2) g2: Create & display basic objects](https://trello.com/c/61L8sQPg/5-2-g2-create-display-basic-objects)

HealthcareSystem/GraphicalEditor/Enumerations/MapObjectDepartments.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace GraphicalEditor.Enumerations
 {
-    public enum MapObjectDepartment
+    public enum MapObjectDepartments
```
vlaksi: Check for [this](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations) and try to use it.

vlaksi (changes requested):

HealthcareSystem/GraphicalEditor/Enumerations/MapObjectDepartments.cs
VasilijeBursac: I understand your comment and judging by it, the name of the enumeration should be singular type name, but I don't know how I would then name the class whose field is of the type of that enumeration.

VasilijeBursac: I have renamed Enumerations according to C# conventions, so check them out now. 😄

milana-todorovic (dismissed):
Seems alright to me.

HealthcareSystem/GraphicalEditor/Enumerations/MapObjectDepartments.cs
vlaksi: Much better, well done!

vlaksi (dismissed):

milana-todorovic:
Merge develop with this branch to resolve conflicts (and tag us on discord afterwards so we can approve again quickly).

milana-todorovic (dismissed):

milana-todorovic (dismissed):

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MockupObjects.cs
```diff
@@ -197,18 +197,18 @@ public List<MapObject> getAllMapObjects()
...
 
-                double yOfObject = i == 0 ? startYOfDrawing : startYOfDrawing + heightOfObject * i - i * Constants.AllConstants.RECTANGLE_STROKE_THICKNESS;
+                double yOfObject = i == 0 ? startYOfDrawing : startYOfDrawing + heightOfObject * i - i * AllConstants.RECTANGLE_STROKE_THICKNESS;
 
 
                 if (i == 3)
                 {
-                    yOfObject += 80 + i * Constants.AllConstants.RECTANGLE_STROKE_THICKNESS;
+                    yOfObject += 80 + i * AllConstants.RECTANGLE_STROKE_THICKNESS;
                 }
```
vlaksi: What is 'i', give it a more descriptive name if you use it in so much line of code.  Also, why you use 6? Can you make const for that? Why I propose const? Because if someone comes to code and looks, they will now why is 6.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MockupObjects.cs
```diff
@@ -227,18 +227,18 @@ public List<MapObject> getAllMapObjects()
...
```
vlaksi, line 230: Also, why 3?

vlaksi (changes requested):

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MockupObjects.cs
milana-todorovic: I actually kind of disagree with this, since this is a mockup class that's getting deleted relatively soon anyway. Tangling it up with constants will make it harder to delete, and trying too hard to make it clean is really just wasted effort.

VasilijeBursac: @milana-todorovic  Yes, I just used "i" as iterator to draw column of rooms. It doesn't have any special meaning and also each of this objects is saved separately to .txt file and will be loaded from it.

milana-todorovic (dismissed):
This last commit should've been marked as merge instead of fix though, fix is for bugs.

VasilijeBursac:
> This last commit should've been marked as merge instead of fix though, fix is for bugs.

Sorry, I will keep that in mind and correct that in future commits.

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MockupObjects.cs
vlaksi: I actually have one question, are you will keep this code for a possible move to Mockup data to test or you will delete it when you finish with it? 

  - If you have an idea to keep it in our code, then I suggest applying my proposal.
  - If you will delete it, and never again use it, that is okay, then make it like this

vlaksi (approved):

milana-todorovic (approved):

HealthcareSystem/GraphicalEditor/Models/MapObjectRelated/MockupObjects.cs
VasilijeBursac: @vlaksi  Okay. We only keep that function in case of some failure with .txt file (to fill the .txt with map objects). It's not constantly used, but just in case if something goes wrong. We will probably delete it in future, but for now on it will stay as a precaution...

PR #27, opened 2020-11-06
Feature g8 display all information

ogikuzma:
Displaying map object information in the right panel and enabling its editing.


VasilijeBursac:
[(2) g8: Display all information (description, working hours,...)](https://trello.com/c/4uXEv66o/8-2-g8-display-all-information-description-working-hours)

HealthcareSystem/GraphicalEditor/Repository/FileRepository.cs
```diff
@@ -27,6 +27,7 @@ public void SaveMap(List<MapObject> allMapObjects)
             serializer.TypeNameHandling = TypeNameHandling.All;
             serializer.Formatting = Formatting.Indented;
             serializer.ContractResolver = new ContractResolver();
+            serializer.ReferenceLoopHandling = ReferenceLoopHandling.Ignore;
```
milana-todorovic, line 30: Also, I assume that some of the changes here were meant to be some sort of fix for the serialization & deserialization of the map? Not sure if you know but that still doesn't work (at least when I try it, deserialization fails after the map is serialized), so I suggest making a fix branch for that sometime in the future.

anciz (changes requested):
Why is function 'Change_Display_Information(object sender, RoutedEventArgs e)' left empty?

milana-todorovic (dismissed):
This is an okay start for the edit panel, but I feel like having some sort of EditModeActive flag in the code behind and binding visibility to that instead of the toggle button would be a better and simpler approach. You could use that to reduce the number of visibility bindings (have one stack panel for editing mode, one stack panel for viewing mode, and just set visibility of those instead of individual elements), which would make it easier to add new elements (a Cancel button for example). Another issue that I see happening is this situation - I select one element, start editing it, and then just select another element instead of finishing the editing first. It would make sense to show the new element's information in view mode instead of in edit mode, or even to warn me and prevent me from switching to a new element if I made changes I haven't saved, but if you're just using a toggle button instead of code behind to track which mode you're in this would be very inconvenient to do.

HealthcareSystem/GraphicalEditor/Repository/FileRepository.cs
ogikuzma: The suggestion you mentioned for the EditModeActive flag is definitely approach I was planning to implement. I've had small problems with that approach and my colleague wanted to start implementing edit functionality, so I left it with this. Refactoring will probably be neccessary when we start the next sprint. Making two stack panels is a good suggestion. I will definetely consider it.
Editing won't look like that. Changes to a field won't be lost by focusing on another field. Only way to discard the changes is by clicking the Cancel button that will be implemented.
As serialization is concerned, I guess the problem is that you do not have a map saved. Try saving map objects from MockupObjects class, commenting that line and then load from file.

ogikuzma:
> Why is function 'Change_Display_Information(object sender, RoutedEventArgs e)' left empty?

It will be implemented as soon as the mouse logic is approved for merging with the develop branch.

anciz (dismissed):

HealthcareSystem/GraphicalEditor/Repository/FileRepository.cs
milana-todorovic: So I tried checking the deserialization issue again, and it's definitely not because there's no map saved, it happens specifically if I create the mockup objects, save them, and then load. It gets kind of odd though, because on this branch I get an exception no matter how I run, but on another branch (feature-g8.1-all-objectinfo-display) it doesn't happen if I run the editor directly, but does if I run from the ManagerFrontend. Here's the exception I'm getting:
image
And the call stack:
image

Could just be something wrong on my copy of the repository, but maybe try reproducing the issue? Also, sorry for clogging up this pull request with this, maybe we should start using issues to report unrelated things we find during code review.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -24,19 +25,43 @@ namespace GraphicalEditor
...
         private IRepository repository;
 
+        private Boolean _editMode = false;
+        public Boolean EditMode
+        {
+            get { return _editMode; }
+            set
+            {
+                _editMode = value;
+                OnPropertyChanged();
```
milana-todorovic: Doesn't OnPropertyChange usually need to fire with the property name as argument? Not sure what happens without it though, if it works like this I assume it doesn't matter.

milana-todorovic (dismissed):
Seems alright.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
ogikuzma: It looked weird for me too, but it works. I will probably add the property name to avoid any possible future problems.

milana-todorovic: Sure, you have conflicts to resolve anyway, so no reason not to add it now.

anciz (dismissed):

HealthcareSystem/GraphicalEditor/Repository/FileRepository.cs
ogikuzma: I am not quite sure what is going on here. I haven't been able to reproduce that exception. This plugin has been producing a lot of weird bugs since I started using it. It seems like I will need to write saving and retrieving from 0 like I initially planned, but the problem is that model is constantly evolving, so I will need to change that class constantly...

ogikuzma:
I should have made two separate commits. Sorry about that.
Exception with the database should be resolved now.

anciz (approved):

milana-todorovic (approved):
Looks good, and I'm not having the deserialization issue anymore. Was edit mode meant to actually work at this point though? Because it currently doesn't really.

ogikuzma:
No, my colleague is working on that feature at this moment.

PR #28, opened 2020-11-06
feat[g8]: Mouse hover and click on map objects

VasilijeBursac:


VasilijeBursac:
[(2) g8: Display all information (description, working hours,...)](https://trello.com/c/4uXEv66o/8-2-g8-display-all-information-description-working-hours)

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -54,6 +59,124 @@ private void LoadMapOnCanvas()
...
+        
+        private MapObject findSelectedMapObject()
+        {
+            List<MapObject> MapObjectsThatContainMouseCursor = findAllMapObjectsThatContainMouseCursor(AllMapObjects);
+
+            MapObject selectedMapObject = findMapObjectWithMinimumArea(MapObjectsThatContainMouseCursor);
+
+            if (selectedMapObject != null)
+            {
+                selectedMapObject.Rectangle.Effect = new DropShadowEffect { Direction = 0, ShadowDepth = 0, BlurRadius = 14, Opacity = 1, Color = Colors.MediumPurple };
```
milana-todorovic, line 85: Having side effects in a method named 'findSelectedMapObjects' isn't very clean. I feel like this method should really just find the object in question, and return it to the caller. Then I would make a method that's something like 'ApplyShadowToSelectedObject' or 'UpdateShadows' and have the Canvas_MouseMove method call that.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -26,6 +27,10 @@ namespace GraphicalEditor
     public partial class MainWindow : Window
     {
         private Canvas _canvas;
+        Point _mouseCurrentPosition;
```
milana-todorovic: Is this field really necessary? The mouse position is already passed to the event handlers, and this field is only used in one method. Why not simply pass the mouse position to the 'findAllMapObjectsThatContainMouseCursor' when it's called instead and remove this field?

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -54,6 +59,124 @@ private void LoadMapOnCanvas()
...
+
+            if (selectedMapObject != null)
+            {
+                selectedMapObject.Rectangle.Effect = new DropShadowEffect { Direction = 0, ShadowDepth = 0, BlurRadius = 14, Opacity = 1, Color = Colors.MediumPurple };
+            }
+
+            return selectedMapObject;
+        }
+
+        private List<MapObject> findAllMapObjectsThatContainMouseCursor(List<MapObject> AllMapObjects)
```
milana-todorovic: Why is this passed to this method when it's already a field of this class? And why is the mouse position used from a field (that no other method needs) instead of being passed here?

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -54,6 +59,124 @@ private void LoadMapOnCanvas()
...
+        {
+            List<MapObject> MapObjectsThatContainMouseCursor = new List<MapObject>();
+
+            foreach (MapObject mapObject in AllMapObjects)
+            {
+                if (isMouseXCoordinateInsideRectangle(_mouseCurrentPosition.X, mapObject.Rectangle) && isMouseYCoordinateInsideRectangle(_mouseCurrentPosition.Y, mapObject.Rectangle))
```
milana-todorovic: This line isn't really readable, why not make a single IsMouseInsideRectangle method that takes the Rectangle and the whole Point instead of just one coordinate?

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -54,6 +59,124 @@ private void LoadMapOnCanvas()
...
+                {
+                    MapObjectsThatContainMouseCursor.Add(mapObject);
+                }
+                else
+                {
+                    mapObject.Rectangle.Effect = null;
```
milana-todorovic: Having a side effect here isn't very clean.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -54,6 +59,124 @@ private void LoadMapOnCanvas()
...
+            {
+                TestText.Text = "";
+            }
+            
+        }
+
+        private void Canvas_MouseUp(object sender, MouseButtonEventArgs e)
+        {
+
+        }
```
vlaksi: Do not leave empty functions.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -54,6 +59,124 @@ private void LoadMapOnCanvas()
...
+            if (MapObjects.Count != 0)
+            {
+                MapObject mapObjectWithMinimumArea = MapObjects[0];
+
+                foreach (MapObject mapObject in MapObjects)
+                {
+                    if ((mapObject.MapObjectArea) < (mapObjectWithMinimumArea.MapObjectArea))
+                    {
+                        mapObjectWithMinimumArea = mapObject;
+                        mapObjectWithMinimumArea.Rectangle.Effect = new DropShadowEffect { Direction = 0, ShadowDepth = 0, BlurRadius = 14, Opacity = 1, Color = Colors.MediumPurple };
```
milana-todorovic: Having side effects here isn't very clean either, the responsibility of this method should be just to find the object with the minimum area. Also, does Linq have a method that finds the minimum of a list based on any property? If it does using that might make things simpler here.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -26,6 +27,10 @@ namespace GraphicalEditor
     public partial class MainWindow : Window
     {
         private Canvas _canvas;
+        Point _mouseCurrentPosition;
+
+        
+
```
vlaksi: You do not need 4 lines of code for this.

vlaksi (dismissed):

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -54,6 +59,124 @@ private void LoadMapOnCanvas()
...
+            }
+            
+        }
+
+
+        private void Canvas_MouseDown(object sender, MouseButtonEventArgs e)
+        {
+            _mouseCurrentPosition = e.GetPosition(this.Canvas);
+
+            MapObject selectedMapObject = findSelectedMapObject();
```
milana-todorovic: This really points out why I have issues with where the logic for applying shadows on hover is placed - for example, the shadow side effects are happening here as well, even though there's no need.

milana-todorovic (changes requested):
I would refactor the code so the logic for adding/removing shadow effects isn't in the methods used for finding the selected objects. It makes more sense to find the selected object first, and then if that's what needs to be done at the moment add the shadow to the selected one, and remove shadows from all the rest. Also, another thing that I forgot to mention, is that it would make sense to have an effect that's applied on hover, and then another effect that's applied after the element is clicked. So for example you have a drop shadow to show the user what element they're hovering on, but then a border or a different colored shadow to show which element is currently selected (and showing in the side panel).

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
VasilijeBursac: This field is mostly used to make code more readable, so that's it's use.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
VasilijeBursac: I could extract that to IsMouseInsideRectangle method but it in it I would do the same thing, because if condition gets very long.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -127,48 +133,28 @@ private MapObject findMapObjectWithMinimumArea(List<MapObject> MapObjects)
...
-            {
-                return false;
-            }
-        }
-
-        private bool isMouseYCoordinateInsideRectangle(double mouseYCoordinate, Rectangle objectRectangle)
+        private bool IsMouseCursorInsideRectangle(Point cursorPosition, Rectangle objectRectangle)
         {
-            if ((mouseYCoordinate > Canvas.GetTop(objectRectangle)) && (mouseYCoordinate < Canvas.GetTop(objectRectangle) + objectRectangle.Height))
+            if ((cursorPosition.X > Canvas.GetLeft(objectRectangle) && cursorPosition.X < Canvas.GetLeft(objectRectangle) + objectRectangle.Width) && ((cursorPosition.Y > Canvas.GetTop(objectRectangle) && cursorPosition.Y < Canvas.GetTop(objectRectangle) + objectRectangle.Height)))
```
milana-todorovic, line 153: Maybe split this into multiple ifs for readability? Something like:
if (cursorPosition.X < Canvas.getLeft(objectRectangle) return false;
if (cursonPositionX > (Canvas.getLeft(objectRectangle) + objectRectangle.Width)) return false;
...
That's not really ideal either though, so I'm not sure what the best way to write this is.

milana-todorovic (approved):
Seems much cleaner now, and looks like it works well on the map.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
VasilijeBursac: Okay, I will take that into consideration, although I prefer it the way I did it, but I agree that it is very questionable what is the best way to refactor it. Thanks for feedback and advices! 😄

vlaksi (approved):

PR #32, opened 2020-11-07
refactor[w1.5]:adding feedback validator

sladjana-savkovic:
Teaching assistent gives a suggestion to create validator layer for checking some conditions.


HealthcareSystem/Backend/Model/Exceptions/HealthClinicException.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Backend.Model.Exceptions
 {
-    class HealthClinicException : Exception
+    public class HealthClinicException : Exception
     {
         public HealthClinicException() : base() { }
         public HealthClinicException(String message) : base(message) { }
```
vlaksi, line 12: For what is used 'HealthClinicException'? An exception should have some descriptive name.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
```diff
@@ -23,17 +23,13 @@ public FeedbackService(IFeedbackRepository feedbackRepository)
         }
         public void AddFeedback(Feedback feedback)
         {
-            if (feedback == null)
-                throw new BadRequestException("Please, write a comment to send feedback.");
             _feedbackRepository.AddFeedback(feedback);
         }
```
vlaksi, line 27: Did you remove exception handling, why?

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
```diff
@@ -23,17 +23,13 @@ public FeedbackService(IFeedbackRepository feedbackRepository)
...
         public void PublishFeedback(int id)
         {
-            Feedback feedback = GetFeedbackById(id);
-            if (!feedback.IsAllowedToPublish)
-                throw new BadRequestException("Feedback is not allowed to publish.");
-            feedback.IsPublished = true;
-            _feedbackRepository.UpdateFeedback(feedback);
+            Feedback _feedback = GetFeedbackById(id);
+            _feedback.IsPublished = true;
+            _feedbackRepository.UpdateFeedback(_feedback);
```
vlaksi: ### First comment

The underscore before a variable name _val is nothing more than a convention. In C#, it is used when defining the **private member variable for public property**.

```sh
private string _val;
public string Values
{
    get { return _val;}
    set {_val = value;}
}
```

So, when I read your code, I expect _feedback is a private member variable for some public property, but that is not true. That is just a local variable in the method. 

### Proposal 

  - use underscore before a variable name only if it is used for defining the **private member variable for public property** (like in the example above)

<br><br>

### Second comment

Why you remove exception handling?

### Proposal

In one layer make throwing exception, and on another make try-catch. For example:
  - in repository layer make throw exception in some cases
  - in-service make try-catch for calling some of those repositories

#### Why is this good?

When we have that logic in our application, we can easily add loggers and logging in to our infrastructure.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
```diff
@@ -45,7 +41,10 @@ public List<Feedback> GetUnpublishedFeedbacks()
         }
         public Feedback GetFeedbackById(int id)
         {
-            return _feedbackRepository.GetFeedbackById(id);
+            Feedback _feedback = _feedbackRepository.GetFeedbackById(id);
+            if (_feedback == null)
+                throw new NotFoundException("Feeback with id=" + id + " doesn't exist in database.");
+            return _feedback;
         }
```
vlaksi: Apply here also what I said in the comment above.

HealthcareSystem/PatientWebApp/Controllers/FeedbackController.cs
```diff
@@ -39,60 +44,51 @@ public ActionResult AddFeedback(FeedbackDTO feedbackDTO)
...
             }
+            Feedback feedback = FeedbackAdapter.FeedbackDTOToFeedback(feedbackDTO);
+            _feedbackService.AddFeedback(feedback);
+            return Ok();
 
         }
 
         [HttpGet("published-feedbacks")]
         public ActionResult getPublishedFeedbacks()
         {
```
vlaksi: ### Comment

General what is problem in FeedbackController class is that you are not consistent.

```sh
public IActionResult GetFeedbackById(int id)
public ActionResult AddFeedback(FeedbackDTO feedbackDTO){}
public ActionResult getUnpublishedFeedbacks(){}
public ActionResult PublishFeedback(int id){}
public ActionResult getPublishedFeedbacks(){}
```

### Proposal

Decide will you use a small beginning letter for the method or big. I propose big, that is a convention in C#.

HealthcareSystem/PatientWebApp/Validators/FeedbackValidator.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            if (feedbackDTO == null)
+                throw new BadRequestException("Feedback cannot be null.");
+            if (feedbackDTO.Comment == null || feedbackDTO.Comment == "")
+                throw new BadRequestException("Comment is required field.");
+            if (feedbackDTO.CommentatorName == null)
+                throw new BadRequestException("Commentator name cannot be null.");
+            if (feedbackDTO.CommentatorSurname == null)
+                throw new BadRequestException("Commentator surname cannot be null.");
+        }
```
vlaksi: ### Comment

Is BadRequestException good for validation?

### Proposal

You could make something like this and then throw that exception.

```sh
namespace Backend.Model.Exceptions
{
    public class ValidationException : HealthClinicException
    {
        public ValidationException () : base() { }
        public ValidationException (string message) : base(message) { }
    }
}
```

HealthcareSystem/PatientWebApp/Validators/FeedbackValidator.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public void checkIfFeedbacksIsAllowedToPublish(int id)
+        {
+            Feedback _feedback = _feedbackService.GetFeedbackById(id);
+            if (!_feedback.IsAllowedToPublish)
+                throw new BadRequestException("Feedback is not allowed to publish.");
+        }
```
vlaksi: The same problem with underscores like in the comments above, try to apply the proposal here also.

vlaksi (changes requested):

vlaksi:
One major problem with this branch generally, is it really refactoring? Or is it adding a new feature (validator) 😅

HealthcareSystem/Backend/Model/Exceptions/HealthClinicException.cs
sladjana-savkovic: This class is like "general" exception at the application level. This is the way for creating own exceptions which was explained by the professor and teaching assistants. In the case of name of this class, it was a suggestion your colleagues from DevOps, and I think the name fully corresponds to the purpose.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
sladjana-savkovic: I didn't remove that exception handling, I just switch it to FeedbackValidation class.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
sladjana-savkovic: Our teaching assistant requires that try-catch be made in controllers layer such as FeedbackController (WebApp).  I suggest that you better familiarize with code, in case you don't see throwing exception and making try-catch blocks.

sladjana-savkovic:
> One major problem with this branch generally, is it really refactoring? Or is it adding a new feature (validator) 😅

I would say it's refactoring, because it was proposed for the first time today at a meeting with an assistant. That was an advice how to improve our code.

milana-todorovic (changes requested):
This change is good and I like the idea of adding a validator class a lot, but I have one suggestion. I feel like the validation logic should be moved slightly. Validation is an integral part of the business logic, so I feel like it should be in the Backend project, not the PatientWebApp, and that it should be called by the service layer, not by the ASP controllers. It doesn't matter that much right now, but it will become apparent once the Backend and PatientWebApp are separated properly and start communicating via a protocol. It shouldn't be left up to the ASP app (which will eventually be a client for the backend) to call the validation or not.
(Of course, let me know if your TA disagrees with me and specifically said you should do it the way it currently is though, so I can switch from Changes requested to Approve.)

milana-todorovic:
> > One major problem with this branch generally, is it really refactoring? Or is it adding a new feature (validator) 😅
> 
> I would say it's refactoring, because it was proposed for the first time today at a meeting with an assistant. That was an advice how to improve our code.

It's definitely refactoring, the validation logic existed already, it's just been moved to a separate class.

HealthcareSystem/Backend/Model/Exceptions/HealthClinicException.cs
milana-todorovic: I think it's definitely a descriptive name, since the responsibility of the class is just to act as the root of the exception hierarchy - thus, it's 'SystemName' + 'Exception'. Of course, you can't really tell when it's meant to be thrown from the name, because it's not meant to be thrown at all.

sladjana-savkovic:
> It doesn't matter that much right now, but it will become apparent once the Backend and PatientWebApp are separated properly and start communicating via a protocol. It shouldn't be left up to the ASP app (which will eventually be a client for the backend) to call the validation or not.
> (Of course, let me know if your TA disagrees with me and specifically said you should do it the way it currently is though, so I can switch from Changes requested to Approve.)

I agree with you completely, but I am not sure that our teaching assistant wants the same thing. He says that validation layer is the connection between controller and adapter layers. Validation is used for verification DTO class before it converts in regular class (such as Feedback). In case of a change a business logic in future, I will move this class in Backend project.

HealthcareSystem/Backend/Model/Exceptions/HealthClinicException.cs
vlaksi: Alright, then it is okay, I am wondering what the usage of this is, now I understand that is the root of the exception hierarchy. Thanks.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
vlaksi: You moved exception handling to layer up, now only exceptions work for your web app?

sladjana-savkovic: Exactly.

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
vlaksi: I have already familiarized myself with code. However, there is something that I disagree with. I am not satisfied because you made exception handling only on your app, I think it is better to have that in our backend maybe. But if your product owner wants that in the controller layer. Alright then, but that is only good for your team I think, not for our organization.

vlaksi:
> > One major problem with this branch generally, is it really refactoring? Or is it adding a new feature (validator) 😅
> 
> I would say it's refactoring because it was proposed for the first time today at a meeting with an assistant. That was advice on how to improve our code.

You are right, you already have that and now you moved it just to separated logic. Seems alright to me then.



> > It doesn't matter that much right now, but it will become apparent once the Backend and PatientWebApp are separated properly and start communicating via a protocol. It shouldn't be left up to the ASP app (which will eventually be a client for the backend) to call the validation or not.
> > (Of course, let me know if your TA disagrees with me and specifically said you should do it the way it currently is though, so I can switch from Changes requested to Approve.)
> 
> I agree with you completely, but I am not sure that our teaching assistant wants the same thing. He says that validation layer is the connection between controller and adapter layers. Validation is used for verification DTO class before it converts in regular class (such as Feedback). In case of a change a business logic in future, I will move this class in Backend project.

I am not sure why your product owner wants that in web-app instead of in our backend infrastructure, could you ask him why?

HealthcareSystem/Backend/Service/NotificationSurveyAndFeedback/FeedbackService.cs
sladjana-savkovic: Our product owner thinks that all exceptions should exist in "higher" layers (such as Validation class in WebApp) because it will prevent that errors come to lower layers. That is, in the case of exception, some message will be returned to the user immediately instead of trying to write "something wrong" to the database. After the first sprint, we can consider your attitude about validation with the product owner.

vlaksi: That is definitely a healthy discussion, and we should talk about it in the next meeting.

sladjana-savkovic:
> > > One major problem with this branch generally, is it really refactoring? Or is it adding a new feature (validator) 😅
> > 
> > 
> > I would say it's refactoring because it was proposed for the first time today at a meeting with an assistant. That was advice on how to improve our code.
> 
> You are right, you already have that and now you moved it just to separated logic. Seems alright to me then.
> 
> > > It doesn't matter that much right now, but it will become apparent once the Backend and PatientWebApp are separated properly and start communicating via a protocol. It shouldn't be left up to the ASP app (which will eventually be a client for the backend) to call the validation or not.
> > > (Of course, let me know if your TA disagrees with me and specifically said you should do it the way it currently is though, so I can switch from Changes requested to Approve.)
> > 
> > 
> > I agree with you completely, but I am not sure that our teaching assistant wants the same thing. He says that validation layer is the connection between controller and adapter layers. Validation is used for verification DTO class before it converts in regular class (such as Feedback). In case of a change a business logic in future, I will move this class in Backend project.
> 
> I am not sure why your product owner wants that in web-app instead of in our backend infrastructure, could you ask him why?

We already asked him today, I wrote an explanation in the comments above.

milana-todorovic:
> > It doesn't matter that much right now, but it will become apparent once the Backend and PatientWebApp are separated properly and start communicating via a protocol. It shouldn't be left up to the ASP app (which will eventually be a client for the backend) to call the validation or not.
> > (Of course, let me know if your TA disagrees with me and specifically said you should do it the way it currently is though, so I can switch from Changes requested to Approve.)
> 
> I agree with you completely, but I am not sure that our teaching assistant wants the same thing. He says that validation layer is the connection between controller and adapter layers. Validation is used for verification DTO class before it converts in regular class (such as Feedback). In case of a change a business logic in future, I will move this class in Backend project.

Did he mention what that validator is supposed to check for? One thing that would make a lot of sense to me is to actually have multiple validators. One of them would end up between the DTO and the model, but it would only really check if a valid conversion is possible - that is, if all the required information is there and in the correct format (without any kind of database access). Meanwhile, validation rules that heavily depend on system state and business logic would stay in the service layer. So the duty of the first validator that's called by the controller or adapter would be to check if a request is made correctly as far as format goes, and the duty of the second validator would be to check if the request (which is already in the correct format) can actually be executed.
However, while I feel like this kind of separation of concerns might be a good idea in the long term, I don't really insist on implementing it right now. It might warrant further consideration and improvements (for example, looking for a more flexible way to write validation rules sometime in the future might be a good idea).

milana-todorovic (dismissed):
Potential improvements were suggested in the comments above, but I think leaving it as is for now is perfectly fine.

sladjana-savkovic:
> > > It doesn't matter that much right now, but it will become apparent once the Backend and PatientWebApp are separated properly and start communicating via a protocol. It shouldn't be left up to the ASP app (which will eventually be a client for the backend) to call the validation or not.
> > > (Of course, let me know if your TA disagrees with me and specifically said you should do it the way it currently is though, so I can switch from Changes requested to Approve.)
> > 
> > 
> > I agree with you completely, but I am not sure that our teaching assistant wants the same thing. He says that validation layer is the connection between controller and adapter layers. Validation is used for verification DTO class before it converts in regular class (such as Feedback). In case of a change a business logic in future, I will move this class in Backend project.
> 
> Did he mention what that validator is supposed to check for? One thing that would make a lot of sense to me is to actually have multiple validators. One of them would end up between the DTO and the model, but it would only really check if a valid conversion is possible - that is, if all the required information is there and in the correct format (without any kind of database access). Meanwhile, validation rules that heavily depend on system state and business logic would stay in the service layer. So the duty of the first validator that's called by the controller or adapter would be to check if a request is made correctly as far as format goes, and the duty of the second validator would be to check if the request (which is already in the correct format) can actually be executed.
> However, while I feel like this kind of separation of concerns might be a good idea in the long term, I don't really insist on implementing it right now. It might warrant further consideration and improvements (for example, looking for a more flexible way to write validation rules sometime in the future might be a good idea).

Just like that, first validator used for check if conversion between DTO and model is possible. The second validator doesn't taken into 
consideration at the last meeting with product owner. I think that he saw no need for it, but that may be the subject of some subsequent sprint.

HealthcareSystem/PatientWebApp/Validators/FeedbackValidator.cs
vlaksi: @sladjana-savkovic just apply this, and then we can merge this branch to develop

vlaksi (approved):
lgtm

sladjana-savkovic:

milana-todorovic (approved):

PR #43, opened 2020-11-18
refactor[w2.1]:refactoring classes about patient and patientCard

sladjana-savkovic:
This pull request is about refactoring backend code about patient and patient card. Class "PatientCard" is connected with class "Examination" which will refactor my colleagues. Because of that, I still can't make new migration. I moved some classes in other packages or changed fields names, that's reason why this commit is so big.


draganacarapic (dismissed):

sladjana-savkovic:

draganacarapic (dismissed):

milana-todorovic (changes requested):
One issue that I notice is that generally speaking, deleting patients isn't a use case that exists (if I remember correctly patient history needs to be kept for ~10 years?), and there's a lot of code here to support that nonexistent use case (such as the repositories for deleted patients and patient cards). I think deleting it will make future work easier.

Other than that, things seem generally ok. Having Users as a package seems logical to me, but I'm less sure about Enums. I feel like putting enums together with objects that use them would make more sense than putting them in a separate package, but I'm ok with leaving the package structure like this for now though, it's an improvement compared to how it was previously. 

Also, I'm not sure if the refactored code should be getting some tests at this point (or if this part doesn't really have anything to really test right now)?

milana-todorovic (changes requested):
Generally looks good, but you forgot to delete this line so I'm getting build issues:
image

Also, could you merge develop to this branch to make it up to date?

milana-todorovic (approved):
Everything seems alright.

draganacarapic (approved):

PR #45, opened 2020-11-19
gui[g-all]: Implemented MaterialDesign, redesigned buttons, colours and styles

VasilijeBursac:
-On this branch I have implemented MaterialDesign in GraphicalEditor project, DoctorWPF (it was necessary because it calls GraphicalEditor project) and updated MaterialDesign version in Manager and Secretary WPFs (this was necessary too, because versions in GraphicalEditor and other WPFs need to match because these WPFs call GraphicalEditor). 
-I have updated colours, button design and style of some controls

This is only XAML code so there is not much to review...

### DIFF
--- a/HealthcareSystem/DoctorFrontend/App.xaml
+++ b/HealthcareSystem/DoctorFrontend/App.xaml
@@ -4,6 +4,23 @@
              xmlns:local="clr-namespace:WpfApp1"
              StartupUri="Wizard/Window1.xaml">
     <Application.Resources>
-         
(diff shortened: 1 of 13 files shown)


ogikuzma (approved):
Everything seems to be ok. One thing I've just noticed is that when you pick a second floor and cancel changes made to the object, the map renders the first floor instead of staying at the second. We will need to investigate this furthermore.

vlaksi (approved):
lgtm

PR #46, opened 2020-11-19
refactor[w2.4]: refactoring Therapy and Examination in all layers

draganacarapic:
This commit is huge because I also changed the files in WPF applications created due to refactoring. Also 
I couldn't separate the Examination and Therapy class in two commits because they are very connected.


sladjana-savkovic (dismissed):

sladjana-savkovic (approved):

milana-todorovic (approved):
Seems alright for now, although some parts of the design look like they might get moved around some more later once the business logic concerning Examinations is overhauled.

Also, just a note, there are build errors in the WPF frontends, but since there's a branch active to fix the issue of Backend refactoring breaking the frontends it's not a problem.

PR #47, opened 2020-11-19
refactor[w2.1]:adding country class and refactor relationships

sladjana-savkovic:
This pull request contains commits about adding new class "Country" in model and update relationships between Country and City.


sladjana-savkovic:

jelenabudisa (approved):

vlaksi (approved):
Very clear and clean.

PR #51, opened 2020-11-19
feat[w2.1]:creating dto, mapper and validator for patient

sladjana-savkovic:
I have a goal to create all classes in Web app which are needed for patient's controller and communication between backend and frontend. These classes include DTO, Mapper and Validator layers. Also, I fix some errors or add some method in backend code (from SIMS).


jelenazeljko55 (approved):

sladjana-savkovic:

HealthcareSystem/PatientWebApp/Validators/PatientValidator.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace PatientWebApp.Validators
+{
+    public class PatientValidator
+    {
+        public PatientValidator() { }
+
+        public void validatePatientFields(PatientDTO patientDTO)
```
vlaksi, line 17: Maybe should be good to make these small if(){..} to little private methods? That would be descriptive and cleaner.

image

vlaksi (approved):

HealthcareSystem/PatientWebApp/Validators/PatientValidator.cs
sladjana-savkovic: Ok, I'll create private methods when I'm sure which checks I still need. For now, I will merge this branch with develop.

PR #52, opened 2020-11-19
refactor[w2.4]:refactoring backend code

jelenabudisa:
This pull request is for refactoring backend code for classes Drug,DrugType,Ingridient,Room and RenovationPeriod in all layers. These classes are needed for user story number w2.4. Pattern used while refactoring is dependency pattern.


jelenabudisa:
[(8) W2.4 As a patient I want to search documentation about my recipes and examinations reports so that I could find information I need.](https://trello.com/c/bKGyskUq/23-8-w24-as-a-patient-i-want-to-search-documentation-about-my-recipes-and-examinations-reports-so-that-i-could-find-information-i-n)

sladjana-savkovic (dismissed):

HealthcareSystem/Backend/Model/MyDbContext.cs
```diff
@@ -17,6 +18,9 @@ public class MyDbContext : DbContext
         public DbSet<Admin> Admins { get; set; }
         public DbSet<Feedback> Feedbacks { get; set; }
         public DbSet<PatientCard> PatientCards { get; set; }
+        public DbSet<Ingredient> Ingridients { get; set; }
+        public DbSet<Drug> Drugs { get; set; }
+        public DbSet<DrugType> DrugTypes { get; set; }
```
milana-todorovic, line 23: The Ingredient and DrugType classes don't have keys as far as I can see, can Entity Framework generate tables for them?

HealthcareSystem/Backend/Model/Manager/RenovationPeriod.cs
```diff
@@ -5,43 +5,46 @@
  ***********************************************************************/
 
 using System;
+using System.ComponentModel.DataAnnotations.Schema;
 
 namespace Model.Manager
 {
     public class RenovationPeriod
```
milana-todorovic, line 12: Does this have a key?

HealthcareSystem/Backend/Model/Manager/Room.cs
```diff
@@ -6,11 +6,13 @@
...
+using System.ComponentModel.DataAnnotations.Schema;
 
 namespace Model.Manager
 {
-   public class Room
-   {
+    public class Room
+    {
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public int Number { get; set; }
```
milana-todorovic: Does this need to be annotated with [Key]? I think Entity Framework only infers keys without the annotation if they're named Id.

HealthcareSystem/Backend/Repository/RoomRepository/MySqlRoomRepository/MySqlRoomRepository.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private readonly MyDbContext _context;
+        public MySqlRoomRepository(MyDbContext context)
+        {
+            _context = context;
+        }
+
+        public void DeleteRoom(int number)
+        {
+            _context.Rooms.Remove((Room)_context.Drugs.Where(d => d.Id == number));
```
milana-todorovic: What does this have to do with Drugs?

HealthcareSystem/Backend/Controller/PlacementInARoomAndRenovationPeriod/RenovationPeriodController.cs
```diff
@@ -6,33 +6,34 @@
...
+            public void AddRenovationPeriod(Model.Manager.RenovationPeriod renovationPeriod)
+            {
+                _renovationPeriodService.AddRenovationPeriod(renovationPeriod);
+            }
 
-        public List<RenovationPeriod> ViewRenovations()
-        {
-            return renovationPeriodService.ViewRenovations();
-        }
+            public void DeleteRenovationPeriod(int roomNumber)
```
milana-todorovic: Why rename Cancel to Delete here? It think Cancel was a more logical name for this in the Controller and Service, calling Delete on the Repository is just the way it's currently implemented? (I'm aware it doesn't matter right now anyway since this part of the system will eventually get overhauled by GraphicalEditor anyway, but it just seems like an unnecessary change. Similarly, I'm actually not sure why it was necessary for you to add RenovationPeriod to the database right now anyway? Do you actually need it for anything right now?)

milana-todorovic (changes requested):
@sladjana-savkovic Can you confirm on the comments about keys whether these are actual issues or not?

HealthcareSystem/Backend/Model/Manager/RenovationPeriod.cs
jelenabudisa: It should have, i left it out

HealthcareSystem/Backend/Model/Manager/Room.cs
jelenabudisa: It is true, i will rename it to be Id

HealthcareSystem/Backend/Repository/RoomRepository/MySqlRoomRepository/MySqlRoomRepository.cs
jelenabudisa: Nothing, it was a mistake

HealthcareSystem/Backend/Controller/PlacementInARoomAndRenovationPeriod/RenovationPeriodController.cs
jelenabudisa: RenovationPeriod  is connected with Room so I had to refactor it. I renamed it to be Delete in all layers to be consistent, but if you think that it is better to leave Cancel in Controller and Service I will do so.

milana-todorovic (approved):
Seems alright, the changes I requested were implemented.

sladjana-savkovic (approved):

PR #55, opened 2020-11-20
feat[w2.1]:adding controller and handling exception

sladjana-savkovic:
- I have a goal to create patient controller for adding new patient in database and reading it.
- I changed some methods in backend because of clean code or handling exceptions.
- Also, I added new class for handling exception, called "DatabaseException", which has a task to indicate on errors about updating data in the database.
- In last sprint, I forgot to add exceptions about reading published/unpublished feedback from database and I have just done it.


sladjana-savkovic:

anciz (dismissed):
The description of what you did is quite good since you helped me to get introduced better with the code you wrote.
I do not like the number of changed files in only one commit, but I know it is sometimes impossible to reduce. 
Why did not you require code review from the others of your team? This time, it is OK, but for future you should get your PR approved by other members of company, not only from us.
Your code is very clean so you did good job! 
Well done.

anciz (approved):

jelenabudisa (approved):

PR #56, opened 2020-11-20
feat[g8]: Selection object colour change effect on mouse click

VasilijeBursac:
This branch contains code for Selection Colour Change Effect on mouse click. When you click on object, it will change colour, so you can always see which object is selected. 

- All of this code related to mouse click will be extracted from MainWindow to separate class very soon
- This is code related to frontend, so it is not the cleanest code ever, but I tried hard to make it as clean as possible
- There is a bug which is not related to this code or feature, but because of it, mouse hover looks weird on two objects. I will fix that bug on different branch


VasilijeBursac:
[(2) g8: Display all information (description, working hours,...): 6h [3]](https://trello.com/c/4uXEv66o/8-2-g8-display-all-information-description-working-hours-6h-3)

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -218,17 +218,19 @@ private void Cancel_Editing_Mode(object sender, RoutedEventArgs e)
 
         private void Canvas_MouseMove(object sender, MouseEventArgs e)
         {
-            MapObject selectedMapObject = FindSelectedMapObject(e.GetPosition(this.Canvas));
-            ApplyShadowEffectToObject(selectedMapObject);
+            MapObject hoverMapObject = FindSelectedMapObject(e.GetPosition(this.Canvas));
```
milana-todorovic, line 221: I like this rename a lot, it's small but makes things a lot more readable.

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -239,33 +241,93 @@ private void Canvas_MouseDown(object sender, MouseButtonEventArgs e)
...
+                    if (!mapObject.Equals(_selectedMapObject))
+                    {
+                        RemoveShadowFromObjectAndDecreaseZIndex(mapObject);
+                    }
+                }
+            }
+        }
+
+
+        private void ApplySelectionEffectToObject(MapObject selectedMapObject)
```
milana-todorovic, line 264: One simple way to make this cleaner (as well as ApplyHoverEffectToObject) would be to first remove the effect from all objects (since you're doing that anyway, whether or not  the object you're applying to is null), and then check if the argument is null, and if it isn't, apply the effect to it. I don't insist on making this change right now though, I'm not sure if that one moment when no object has the effect applied will make an odd visible twitch when the change is happening.

milana-todorovic (dismissed):
I like this change a lot, it's a small thing but it'll really improve experience for anyone using the map. And I actually don't have many complaints about clean code here, the suggestion from the comment I wrote could improve DRY, but the methods are pretty well separated and named so the code is fairly readable.

Also, what objects have hover issues? I don't see the bug.

VasilijeBursac:
> I like this change a lot, it's a small thing but it'll really improve experience for anyone using the map. And I actually don't have many complaints about clean code here, the suggestion from the comment I wrote could improve DRY, but the methods are pretty well separated and named so the code is fairly readable.
> 
> Also, what objects have hover issues? I don't see the bug.

Thank you! Hover bug just happens on two objects like a little bit thicker line, but its not rally noticeable, it's just aesthetics... Anyway that is fixed on the next PR...

HealthcareSystem/GraphicalEditor/MainWindow.xaml.cs
VasilijeBursac: Good point. I think that I have successfully implemented your advice in my new commit. I have tested it for awhile and it seems to be working fine. Check it out. 😄

milana-todorovic (commented):
Seems good, just merge develop to this branch just in case since it's behind by a lot.

VasilijeBursac:
> Seems good, just merge develop to this branch just in case since it's behind by a lot.

I did it now. Sorry, it didn't allow me to commit and put name according to our conventions. Even though I made commit (but there was nothing to commit) and wrote commit message it just pushed with this message...

milana-todorovic (approved):

anciz (approved):

PR #57, opened 2020-11-20
fix[g4]: Loading mockup objects - Fixed error in loops for drawing objects

VasilijeBursac:
- This branch contains code for drawing all map objects.
- We used MockupObjects class to load initial objects to file, but there was bug in for loop for drawing column of rooms, which caused two rooms two load twice instead of once. Because of this, json file from which we load map objects also had this problem. 
- This bug caused another bug for Selection Effect on mouse click on those two objects, so this fix solved that bug, too.


VasilijeBursac:
[(5) g4: Build a map from basic objects: 4h [2]](https://trello.com/c/BiuN58fv/21-5-g4-build-a-map-from-basic-objects-4h-2)

milana-todorovic (approved):
Seems alright.

anciz (approved):

PR #60, opened 2020-11-20
feat[w2.4]:creating test for get patient method

sladjana-savkovic:
- I have a goal to create unit tests for patient controller methods.
- Actually, this branch refers to W2.3 user story (test for getPatientByJmbg method) and W2.1 user story (test for addPatient method).
- I accidentally made a mistake about commit message (write 2.4 instead 2.3 user story).


sladjana-savkovic:
[(3) W2.3 As a patient I want to observe my profile information so that I could check if my data is correct.](https://trello.com/c/CbUobywv/22-3-w23-as-a-patient-i-want-to-observe-my-profile-information-so-that-i-could-check-if-my-data-is-correct)

HealthcareSystem/Backend/Model/Users/City.cs
```diff
@@ -24,15 +24,36 @@ public class City
...
+                Country = new Country(city.Country);
+                CountryId = city.Country.Id;
+            }
+            else
+            {
+                Country = new Country();
+                CountryId = 0;
+            }
         }
     }
```
vlaksi: This is good, just one suggestion to stop DRY, try to extract method something like 'InitCountry' or some better description for it. And then use it in constructors.

```sh
        public City(int zipCode,string name,Country country)
        {
            ZipCode = zipCode;
            Name = name;
            InitCountry(country);
        }

        public City(City city)
        {
            ZipCode = city.ZipCode;
            Name = city.Name;
            InitCountry(city.Country);

        }

        private void InitCountry(Country country)
        {
            if (country != null)
            {
                Country = new Country(country);
                CountryId = country.Id;
            }
            else
            {
                Country = new Country();
                CountryId = 0;
            }
        }

```

HealthcareSystem/PatientWebApp/DTOs/PatientDTO.cs
```diff
@@ -29,5 +29,31 @@ public class PatientDTO
         public string Password { get; set; }
 
         public PatientDTO() { }
+
+        public PatientDTO(string name, string surname, string jmbg, GenderType gender, DateTime dateOfBirth, 
+                          string phone, int countryId, string countryName, int cityZipCode, string cityName, 
+                          string homeAddress, BloodType bloodType, RhFactorType rhFactor, bool hasInsurance, 
+                          string lbo, string alergies, string medicalHistory, string email, string password)
+        {
```
vlaksi, line 37: I am not sure really about this, is good to have too many parameters in DTO? Or maybe make some structure for some of these parameters?

HealthcareSystem/PatientWebAppUnitTests/PatientControllerTests.cs
```diff
@@ -0,0 +1,117 @@
...
+using System;
+using Xunit;
+using Model.Enums;
+using Microsoft.AspNetCore.Mvc;
+using Backend;
+using System.Collections.Generic;
+using PatientWebApp.DTOs;
+
+namespace PatientWebAppUnitTests
+{
```
vlaksi: Is a good practice to use:
  - ProjectNameUnitTests
  - ProjectName.UnitTest
  - ProjectName.Tests

One is definitely, you make good prefix, **PatientWebApp**. We just need to make a convention in our organization. Convention about, **how to give a name to test projects**. And after that, everyone else should use that convention.

### My suggestion

Make this name for the project: **PatientWebApp.Tests**, after that, we can have a directory for **Unit tests** and a directory for **integration tests**.

HealthcareSystem/PatientWebAppUnitTests/PatientControllerTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        private PatientDTO CreateInvalidPatientDTOForTestingAddMethod()
+        {
+            PatientDTO patientDTO = new PatientDTO(name: "Ana", surname: "Anic", jmbg: null, gender: GenderType.Z,
+                                                    dateOfBirth: DateTime.Now, phone: "", countryId: 1, countryName: "Srbija",
+                                                    cityZipCode: 21000, cityName: "Novi Sad", homeAddress: null,
+                                                    bloodType: BloodType.A, rhFactor: RhFactorType.Negativna, hasInsurance: false, lbo: "",
+                                                    alergies: null, medicalHistory: null, email: null, password: "");
+            return patientDTO;
+        }
```
vlaksi: Think about **Factory pattern** and try to extract these methods in one class which will do this job, and make the Test class only for tests. In Test class, we only should have **[Fact]** methods. And these methods should call some factory methods.

vlaksi (changes requested):
In general, you make a great job, you bring tests into our infrastructure. That is great! Let's just make some conventions for our organization about that. Also, make little refactoring and we are done with this pr.

HealthcareSystem/PatientWebApp/DTOs/PatientDTO.cs
sladjana-savkovic: I think that PatientDTO must have a lot of parameters because it is one of the requirements when registering a patient. I think that the division into new structures would make confusion and make work more difficult. Especially because all the fields refer to the patient and I don't see meaningful wholes that I could segregate.

vlaksi: We may not have understood each other properly. It's okay to have a lot of parameters. I think to the more readable and reusable structure of code, this too hard to read. The division into new structures would make cleaner and better code, that is all point about clean code.

Look how I propose refactoring in [this pr](https://github.com/PSW-2020-ORG4/HealthcareSystem/pull/61#issuecomment-731607853). I extracted some similar properties, and after that, you can reuse those of structure and validate only them if you want.

I’m not sure if is a practice to use smarter structure in dto's. And if we don't use that in dto's. Then that is okay. But if it is a practice to have some smarter structure in dto's, we should look at [this pr](https://github.com/PSW-2020-ORG4/HealthcareSystem/pull/61#issuecomment-731607853) and make similar to this code.

sladjana-savkovic (commented):
This is implementation about factory pattern, not a method.

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/CreatePatient.cs
```diff
@@ -0,0 +1,27 @@
...
+using Model.Users;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace PatientWebAppTests.CreateObjectsForTests
+{
+    public class CreatePatient : ICreateObject
+    {
+        public object createValidObject()
```
vlaksi: Why **object**, maybe we should use the Patient return type. Do we start our methods with big letters or small? I forget, what is the convention? If we use one convention, then we should use that in all our code.

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/ICreateObject.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace PatientWebAppTests.CreateObjectsForTests
+{
+    public interface ICreateObject
+    {
+        object createValidObject();
+        object createInvalidObject();
```
vlaksi: Why object?

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/ObjectFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+                return new CreatePatient();
+            }
+            if (objectType.Equals("PatientCard"))
+            {
+                return new CreatePatientCard();
+            }
+
+            return null;
+        }
+    }
```
vlaksi: ObjectFactory, which type of object? We should consider a better name here.

HealthcareSystem/PatientWebAppUnitTests/UnitTests/PatientControllerTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            ICreateObject newValidObject = _objectFactory.GetObject("PatientCard");
+            var patientCards = new List<PatientCard>();
+            patientCards.Add((PatientCard)newValidObject.createValidObject());
+
+            patientCardStubRepository.Setup(m => m.GetPatientCardByJmbg("1234567891234")).Returns(patientCards[0]);
+            patientCardStubRepository.Setup(m => m.AddPatientCard(new PatientCard()));
+
+            return patientCardStubRepository.Object;
+
+        }
```
vlaksi: Will you put this in one class or leave it here?

vlaksi (dismissed):

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/ICreateObject.cs
sladjana-savkovic: What return value should I put if I want this factory to create different types of objects? If the solution is to separate factories by objects, then instead of 2 methods for creating valid and invalid objects, I will have 4 classes for each type of object. I think it is overkill.

HealthcareSystem/PatientWebAppUnitTests/UnitTests/PatientControllerTests.cs
sladjana-savkovic: I will leave it here for now.

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/CreatePatient.cs
sladjana-savkovic: I accidentally put the name of the method with small letter.

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/ICreateObject.cs
vlaksi: The method should do one thing, not two, only because of that I suggest separation. But if you think that is not good, okay, this is still your code.

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/CreatePatient.cs
vlaksi: It's okay, I just got confused and asked about the convention.

HealthcareSystem/PatientWebAppUnitTests/UnitTests/PatientControllerTests.cs
vlaksi: You can make in future branch for refactoring this, that is my proposal. As I suppose you don't have time now, and for me that it's okay.

HealthcareSystem/PatientWebAppUnitTests/PatientControllerTests.cs
vlaksi: @sladjana-savkovic what do you think about this?

vlaksi: @PSW-2020-ORG4/developers look at this convention when you make a unit test:
  - ProjectName
  - ProjectNameTest
    - Unit
    - Integration

Like in this image (for now there is the only unit, but when you make integration put them in one directory):

image

HealthcareSystem/PatientWebApp/DTOs/PatientDTO.cs
sladjana-savkovic: I could refactor that in some future pull requests. For now, I would like to merge with develop, so that my team can see my tests and continue with testing.

vlaksi: That it's okay for me.

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/ICreateTestObject.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace PatientWebAppTests.CreateObjectsForTests
+{
+    public interface ICreateTestObject
```
milana-todorovic: I think keeping this interface might be nice, but its real purpose would really be just to make the naming of the methods in the CreateX classes consistent. To do that I would make it generic and return the correct type.

HealthcareSystem/PatientWebAppUnitTests/CreateObjectsForTests/TestObjectFactory.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace PatientWebAppTests.CreateObjectsForTests
+{
+    public class TestObjectFactory
+    {
+        public ICreateTestObject GetObject(string objectType)
```
milana-todorovic: Obviously, if you decide to remove this interface this class goes as well, but I have two suggestions on what to do with this class if the interface is kept and made generic:
1. Keep this class, but split this method into three. This makes sense if you expect to have multiple implementations of ICreateTestObject<PatientDTO>, for example, or if you really just want your tests to not instantiate the creator objects directly.
2. Get rid of this class entirely anyway, and just instantiate the correct creator in your tests directly.

milana-todorovic (changes requested):
I think the tests are okay and definitely a good start (and whether or not the controller converts messages correctly seems like a good thing to test to me). One suggestion for the tests would be to maybe look into making them a Theory instead of Fact for the sake of DRY, but I definitely don't mind if you keep them as is (the 'changes requested' status of this review doesn't refer to this).

The refactoring with the factory went a little overboard though. The only actual clean code concern here was to get the CreateValidX and CreateInvalidX methods out of the PatientControllerTests class. And really, just having the CreatePatient, CreatePatientDTO, and CreatePatientCard classes was enough to do that job, the interface and factory might be a bit overkill (returning object anywhere is a good indicator that you might have an abstraction that doesn't work, I think). I made some suggestions on what to do with the interface and factory in the comments.

milana-todorovic (approved):
Looks good!

draganacarapic (approved):

PR #63, opened 2020-11-21
feat[w2.1]:make ajax call to controller

sladjana-savkovic:
- I have a goal to connect backend and frontend for patient registration.
- This includes : validate fields on frontend, ajax call to controller and upload patient image and save it in memory


vlaksi (approved):
LGTM, but please, next time make more commits with fewer files changed. One commit should do one thing. This **make ajax call to controller** can't be the name of commit and branch, that is an indicator for too big commit.

sladjana-savkovic:

draganacarapic (approved):

PR #66, opened 2020-11-22
feat[w2.2]: sending mail to activate patients account

draganacarapic:
* In this committee I made the logic for sending an email when registering a patient.
* The email contains a button that activates the patient and redirects him to the home page of the web app.
* I also did a test that checks patient activation(update).


draganacarapic:

HealthcareSystem/Backend/Model/Users/WelcomeRequest.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class WelcomeRequest
+    {
+        public string ToEmail { get; set; }
+        public string UserName { get; set; }
+        public string Jmbg { get; set; }
+
+        public WelcomeRequest() { }
+    }
+}
```
vlaksi: Is **WelcomeRequest** a good idea to be placed at Backend->Model->Users directory? Maybe we could make some directory for requests?

HealthcareSystem/Backend/Service/SendingMail/MailService.cs
```diff
@@ -0,0 +1,45 @@
...
+            email.Subject = $"Welcome {request.UserName}";
+            var builder = new BodyBuilder();
+            builder.HtmlBody = MailText;
+            email.Body = builder.ToMessageBody();
+            using var smtp = new SmtpClient();
+            smtp.Connect(_mailSettings.Host, _mailSettings.Port, SecureSocketOptions.StartTls);
+            smtp.Authenticate(_mailSettings.Mail, _mailSettings.Password);
+            await smtp.SendAsync(email);
+            smtp.Disconnect(true);
+        }
```
vlaksi: This method is too long, we could do some refactoring. 

### My proposal for refactoring this class.

```sh
using System.IO;
using System.Threading.Tasks;
using Backend.Model.Users;
using Backend.Settings;
using MailKit.Net.Smtp;
using MailKit.Security;
using Microsoft.Extensions.Options;
using MimeKit;

namespace Backend.Service.SendingMail
{
    public class MailService : IMailService
    {
        private readonly MailSettings _mailSettings;

        public MailService(IOptions<MailSettings> mailSettings)
        {
            _mailSettings = mailSettings.Value;
        }
        public async Task SendWelcomeEmailAsync(WelcomeRequest request)
        {
            var MailText = ParseMailText(request, ReadMailText());
            var email = ConfigureEmail(request, MailText);
            await SendEmail(email);
        }

        private async Task SendEmail(MimeMessage email)
        {
            using var smtp = new SmtpClient();
            smtp.Connect(_mailSettings.Host, _mailSettings.Port, SecureSocketOptions.StartTls);
            smtp.Authenticate(_mailSettings.Mail, _mailSettings.Password);
            await smtp.SendAsync(email);
            smtp.Disconnect(true);
        }

        private MimeMessage ConfigureEmail(WelcomeRequest request, string MailText)
        {
            var email = new MimeMessage();
            email.Sender = MailboxAddress.Parse(_mailSettings.Mail);
            email.To.Add(MailboxAddress.Parse(request.ToEmail));
            email.Subject = $"Welcome {request.UserName}";

            var builder = new BodyBuilder();
            builder.HtmlBody = MailText;
            email.Body = builder.ToMessageBody();
            return email;
        }

        private string ParseMailText(WelcomeRequest request, string MailText)
        {
            var parsedMailText = MailText.Replace("[username]", request.UserName).Replace("[email]", request.ToEmail)
                .Replace("[jmbg]", request.Jmbg);
            return parsedMailText; ;
        }

        private string ReadMailText()
        {
            var pathToHtmlPage = "\\Templates\\WelcomeTemplate.html";
            string FilePath = Directory.GetCurrentDirectory() + pathToHtmlPage;
            StreamReader reader = new StreamReader(FilePath);
            string MailText = reader.ReadToEnd();
            reader.Close();
            return MailText;
        }
    }
}


```

HealthcareSystem/PatientWebApp/Controllers/MailController.cs
```diff
@@ -0,0 +1,43 @@
...
+            _mailService = mailService;
+            _patientValidator = new PatientValidator();
+        }
+
+        [HttpPost("welcome")]
+        public async Task<IActionResult> SendWelcomeMail([FromForm] WelcomeRequest request)
+        {
+            try
+            {
+                 _patientValidator.checkIfEmailValid(request.ToEmail);
```
vlaksi: Rename checkIfEmailValid to CheckIfEmailValid, which is a convention in our organization. Methods should start with capital letters.

vlaksi (dismissed):
In general, good job. Just a couple of suggestions and we are done.

vlaksi:
By accident, I click approve instead of request change. So because of that, I dismissed my approval.

HealthcareSystem/Backend/Service/SendingMail/MailService.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        private readonly MailSettings _mailSettings;
+
+        public MailService(IOptions<MailSettings> mailSettings)
+        {
+            _mailSettings = mailSettings.Value;
+        }
+        public async Task SendWelcomeEmailAsync(WelcomeRequest request)
+        {
+            string FilePath = Directory.GetCurrentDirectory() + "\\Templates\\WelcomeTemplate.html";
```
milana-todorovic, line 26: I really like that you have a template for this for flexibility, but should we maybe pull the path to the templates into app.config? @vlaksi what do you think?

HealthcareSystem/PatientWebApp/Controllers/MailController.cs
```diff
@@ -0,0 +1,43 @@
...
+using Backend.Service.SendingMail;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using PatientWebApp.Validators;
+
+namespace PatientWebApp.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class MailController : ControllerBase
```
milana-todorovic: Basically, I don't think this should be a separate controller. Mail isn't generally something that your application deals with or offers as a resource, so having a path like 'api/mail' is kind of misleading. Instead this is just used during the registration and account activation, so it should be in whichever controller generally deals with accounts.

HealthcareSystem/PatientWebApp/Controllers/MailController.cs
```diff
@@ -0,0 +1,43 @@
...
+        private readonly PatientValidator _patientValidator;
+
+        public MailController(IMailService mailService)
+        {
+            _mailService = mailService;
+            _patientValidator = new PatientValidator();
+        }
+
+        [HttpPost("welcome")]
+        public async Task<IActionResult> SendWelcomeMail([FromForm] WelcomeRequest request)
```
milana-todorovic: I don't think this should be an endpoint - who is meant to call it? If I understood correctly, your use case is this - a new patient fills out the registration form, and clicks on register. The account is created but not yet activated. An activation email is sent automatically to the patient so that they can activate their account. 
If that's the case, I think you should just call this service from whichever method does patient registration. (Quick digression, but that might also be a nice spot for a unit test with the MailService mocked.)

HealthcareSystem/PatientWebApp/Controllers/PatientController.cs
```diff
@@ -71,5 +71,25 @@ public ActionResult AddPatient(PatientDTO patientDTO)
...
             return Ok();
         }
+
+        /// <summary>
+        /// /activate patient (property: IsActive) to true
+        /// </summary>
+        /// <param name="id">id of the object to be changed</param>
+        /// <returns>if alright returns code 200(Ok), if not 400(bed request)</returns>
+        [HttpPut("{jmbg}")]
+        public ActionResult ActivatePatient(string jmbg)
```
milana-todorovic: So this on the other hand definitely does need to be an endpoint, but looking at it, does the path 'api/mail/{jmbg}' look like something you would use for account activation? Wouldn't something like 'api/user/activate/{id}' or 'api/patients/activate/{id}' make more sense? 
Also, I assume you know this and that it isn't implemented because it isn't exactly our concern for this project, but it should really be {horrible encrypted string} instead of {jmbg}. That's absolutely a moot point though, as far as I'm aware we're not even encrypting passwords before sticking them in storage right now so making the activation path this easy to guess is really whatever.

milana-todorovic (commented):
Here are some comments with opinions about your REST API design. They're left as comments instead of a request for changes both because I need you to correct me if I understood your use case wrong, and because I'd mostly just like to hear more opinions on this topic.

HealthcareSystem/PatientWebApp/wwwroot/js/activate_account.js
```diff
@@ -0,0 +1,26 @@
+﻿$(document).ready(function () {
+    $("#activate").click(function (e) {
+        e.preventDefault();
+
+        let url_split = window.location.href.split("?")[1];
+
+        var jmbg = url_split.split("=")[1];
+
+        $.ajax({
```
milana-todorovic: Asking just because you said you didn't want to have the page with the activation button but couldn't immediate redirect to work, what exactly happens if you just put the jmbg extraction and ajax directly into $(document).ready instead of $("#activate").click?

milana-todorovic (commented):
(ignore this, I was going to make another section with comments and clicked sumbit review too early)

HealthcareSystem/PatientWebApp/appsettings.json
```diff
@@ -1,13 +1,20 @@
...
+            "Microsoft": "Warning",
+            "Microsoft.Hosting.Lifetime": "Information"
+        }
+    },
+    "MailSettings": {
+        "Mail": "<fromemail>",
+        "DisplayName": "<displayname>",
+        "Password": "<yourpasswordhere>",
+        "Host": "smtp.gmail.com",
+        "Port": 587
```
milana-todorovic, line 14: I love that you already made this configurable.

HealthcareSystem/Backend/Service/SendingMail/MailService.cs
```diff
@@ -0,0 +1,45 @@
...
+            StreamReader str = new StreamReader(FilePath);
+            string MailText = str.ReadToEnd();
+            str.Close();
+            MailText = MailText.Replace("[username]", request.UserName).Replace("[email]", request.ToEmail).Replace("[jmbg]", request.Jmbg);
```
milana-todorovic: I also love that you made a template for this! I wonder if maybe you should put the path to the template into appsettings.json as well to further improve it. @vlaksi @Pufke @anciz What do you think, is that going to be necessary when we containerize this? I feel like it might be nicer anyway even if we can get it to work correctly as is.

HealthcareSystem/PatientWebApp/Templates/WelcomeTemplate.html
```diff
@@ -0,0 +1,169 @@
...
+                                                        </tr>
+                                                        <!-- Button -->
+                                                        <tr>
+                                                            <td align="center">
+                                                                <table class="center" border="0" cellspacing="0" cellpadding="0" style="text-align: center;">
+                                                                    <tr>
+
+                                                                        <td class="pink-button text-button" style=" background: #006699;color: #c1cddc;font-family: 'Muli', Arial, sans-serif;font-size: 14px; line-height: 18px;padding: 12px 30px;text-align: center;border-radius: 0px 22px 22px 22px;font-weight: bold;">
+
+                                                                            <a href="http://localhost:65117/html/activate_account.html?id=[jmbg]" target="_blank" class="link-white" style="color: #ffffff;text-decoration: none;">Click here </a>
```
milana-todorovic: This path definitely needs to be configurable, otherwise it's going to be an issue when it's time to deploy this. Maybe do something like "[host]/activate_account.html?id=[jmbg]" and either put the host in the config or somehow get it from ASP (I feel like that should be possible?). Or just do "[activation_path]" and assemble the whole path in the MailService.

milana-todorovic (dismissed):
Here are some general comments about making things configurable. The one about the activation path in the template is a change request since it'll cause issues later.

HealthcareSystem/Backend/Service/SendingMail/MailService.cs
vlaksi: That is definitely something I suggest.

draganacarapic:
I was able to change most of the things you suggested to me.
The only problem is that now there is no more MailController, the async function is called in PatientController, so I need to modify the PatientControllerTests class to work with that async function.
That will be my next commit.

HealthcareSystem/PatientWebApp/Controllers/MailController.cs
milana-todorovic: Applied in f3dd33a7ec46560896650fb783d63778066c2722.

HealthcareSystem/PatientWebApp/Controllers/MailController.cs
milana-todorovic: Applied in f3dd33a7ec46560896650fb783d63778066c2722.

HealthcareSystem/PatientWebApp/Controllers/PatientController.cs
milana-todorovic: Correction: the initial path was 'api/patients/{jmbg}', not 'api/mail/{jmbg}' (although the point still stands). Changed to 'api/patients/activate/{jmbg}' in f3dd33a7ec46560896650fb783d63778066c2722.

HealthcareSystem/Backend/Service/SendingMail/MailService.cs
milana-todorovic: Applied as suggested in f3dd33a7ec46560896650fb783d63778066c2722.

HealthcareSystem/PatientWebApp/Controllers/MailController.cs
milana-todorovic: Applied in f3dd33a7ec46560896650fb783d63778066c2722.

sladjana-savkovic (approved):

vlaksi (approved):
lgtm

HealthcareSystem/PatientWebApp/wwwroot/js/activate_account.js
milana-todorovic: This worked, applied in ba19e00e8e1fc56824a0e731412f2cafdc3c5403.

PR #68, opened 2020-11-23
feat[w2.1]:add conditions for registration and refactor

sladjana-savkovic:
- I add some conditions for patient registration and refactor code in Patient Validation class
- I fix bugs about adding  feedback and survey 
- Perhaps a better prefix for this branch is hotfix.


sladjana-savkovic:

HealthcareSystem/PatientWebApp/Validators/PatientValidator.cs
```diff
@@ -15,66 +15,93 @@ public class PatientValidator
...
+            }
+        }
+
+        private bool CheckIfObjectEmpty(PatientDTO patientDTO)
+        {
+            if (patientDTO == null) return true;
+            return false;
+        }
+
+        private bool CheckIfFieldEmpty(string field)
```
milana-todorovic, line 74: Why not just use String.IsNullOrEmpty here? I'm not going to request changes over that but it might be nicer.

HealthcareSystem/Backend/Service/SendingMail/MailService.cs
```diff
@@ -34,6 +34,7 @@ public async Task SendWelcomeEmailAsync(WelcomeRequest request)
         private async Task SendEmail(MimeMessage email)
         {
             using var smtp = new SmtpClient();
+            smtp.CheckCertificateRevocation = false;
```
milana-todorovic, line 37: What was the issue here? Maybe next time you have something like this that isn't really obvious / familiar, write a short explanation of what happened in the commit description or on the PR. It might be useful both to reviewers when reading this, and to you as future reference.

milana-todorovic (approved):
Seems okay, although since it's really just a couple of unrelated fixes / refactors it would be nice if you split up the commits next time. 
As far as the branch prefix goes, I agree, it's not really a feature, but maybe just bugfix would be better than hotfix since none of these seem like emergencies.

HealthcareSystem/Backend/Service/SendingMail/MailService.cs
sladjana-savkovic: [This ](https://stackoverflow.com/questions/59026301/sslhandshakeexception-an-error-occurred-while-attempting-to-establish-an-ssl-or) was a problem.

draganacarapic (approved):

PR #70, opened 2020-11-23
fix[w2.2]: modification of data in the test

draganacarapic:
Due to changes in the Validator class, the tests did not pass (a colleague added a method that prevents the entry of a password of less than 8 characters). So I only changed the data related to the password in the test according to this method.


draganacarapic:

sladjana-savkovic (approved):

vlaksi (approved):
lgtm

PR #71, opened 2020-11-24
gui[g-general]: Implemented Menu for App Functionalities

VasilijeBursac:
- This branch contains code for app main menu. From this menu user will call different functionalities, and different options will be shown in this panel.
- Method for changed menu item is just prepared for future implementation so there is no code in switch case.

This is mostly simple code and XAML code so there is not much to review...


milana-todorovic (approved):
Looks good to me!

anciz (approved):
Nice!

PR #72, opened 2020-11-24
refactor[w2.1]:move uploading image to patient registration page

sladjana-savkovic:
-  I have a goal to add uploading image at existing page for patient registration
- I do this by switching logic from one html page to another


sladjana-savkovic:

vlaksi (approved):
lgtm

draganacarapic (approved):

PR #73, opened 2020-11-24
refactor[ia]:changing project to mvc

Stefan-Kitanovic:
Initially our project was created from a Razor pages template, but we found out that it wouldn't work well with REST controllers, so I have changed entire project to classic ASP.NET MVC instead.


borislav-puzic (dismissed):
lgtm

milana-todorovic (changes requested):
When creating the project, did you uncheck the 'Configure for HTTPS option'? It keeps asking me to install a security certificate when I try to run the project, I think that was supposed to be turned off like the instructions [here](https://www.youtube.com/watch?v=xK0oyVIA7o4&feature=youtu.be&ab_channel=pswftn) say.

Stefan-Kitanovic:
I left it checked while I was creating the project by accident,  I think it should be disabled now.

borislav-puzic (approved):
All good, run project this time and https is off

milana-todorovic (approved):
Seems alright now, I don't have any issues with running the project.

PR #74, opened 2020-11-25
feat[w2.2]: encrypt jmbg in url

draganacarapic:
* I added two new methods for jmbg encryption and decryption, due to the protection of personal data and the impossibility of viewing the patient's jmbg in the url path of the mail activation page,
* Also in the second commit I amended the PatientControllerTests, in order to adhere to the principles of the DRY, in accordance with the professor's suggestion.


HealthcareSystem/PatientWebApp/wwwroot/js/activate_account.js
```diff
@@ -46,4 +48,34 @@
     });
     });
     
-});
\ No newline at end of file
+});
+
+function encryptJmbg(jmbg) {
```
milana-todorovic: Wouldn't this be decryptJmbg? Also, I'm not sure putting the decryption algorithm in javascript is a good idea, maybe sending it still encrypted to the activate endpoint and decrypting it there would be better? I'm definitely not an expert on security though. 
Also, if you do end up also moving decryption to the backend, maybe consider making a small EncryptionService to take over this responsibility.

milana-todorovic (commented):
Leaving this as a comment since I'm not sure what your exact requirement was, to never have the jmbg in urls (including REST endpoints), or just to not have it in the url for the activation page since that can be seen in the browser.

milana-todorovic (commented):
10fe5098a28cde85d73dfaecc9986b915192c31a looks good, although maybe this branch wasn't the best place for it.

HealthcareSystem/PatientWebApp/wwwroot/js/activate_account.js
draganacarapic: Okay, I'll consider moving the decryption to the backend code.

draganacarapic:

HealthcareSystem/Backend/Service/Encryption/EncryptionService.cs
```diff
@@ -0,0 +1,117 @@
...
+            char[] arr;
+            string result = "";
+            arr = encryptedJmbg.ToCharArray(0, 13);
+
+            foreach (char c in arr)
+            {
+                if (c.ToString() == "A")
+                {
+                    result += "0";
+                }
```
anciz, line 24: I am wondering why **'c'** and **'arr'**.
(I do not say it is incorrect, but I am confused)

draganacarapic: arr is abbreviated from array
c is abbreviated from character

anciz (dismissed):
This is well done!

milana-todorovic (dismissed):
Looks good!

milana-todorovic (approved):

Pufke (approved):

PR #75, opened 2020-11-25
feature-ia3.1-action-benefit-service

borislav-puzic:
I want to make service that receives rabbitmq messages that contain actions and benefits, and saves them to database


HealthcareSystem/Backend/Service/ActionBenefitService/RabbitMqActionBenefitService/RabbitMqActionBenefitService.cs
```diff
@@ -0,0 +1,138 @@
...
+            consumer.Registered += OnConsumerRegistered;
+            consumer.Unregistered += OnConsumerUnregistered;
+            consumer.ConsumerCancelled += OnConsumerCancelled;
+
+            _channel.BasicConsume(queue: _queueName, autoAck: false, consumer: consumer);
+
+            return Task.CompletedTask;
+        }
+
+        public void Unsubscribe(String exchangeName)
```
milana-todorovic, line 90: Should there be a Subscribe method as well? I feel like that is probably a use case as well, like when a new Pharmacy is added and subscribed to during runtime?

HealthcareSystem/Backend/Service/ActionBenefitService/RabbitMqActionBenefitService/RabbitMqActionBenefitService.cs
```diff
@@ -0,0 +1,138 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Backend.Service
+{
+    public class RabbitMqActionBenefitMessageingService : BackgroundService
```
milana-todorovic: I would maybe make an interface for this as well, something like IActionBenefitMessageingService? Just to to make the rest of the system a bit less tied to RabbitMQ (like whatever part of the system will call Unsubscribe / Subscribe, I assume it will be the pharmacy service or something - it would also be useful for testing that part of the system). Also, if this will need to be injected anywhere it needs an interface.

HealthcareSystem/Backend/Model/Pharmacies/ActionBenefit.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        [ForeignKey("Pharmacy")]
+        public int PharmacyId { get; set; }
+        public virtual Pharmacy Pharmacy { get; set; }
+
+        [Required]
+        public string Message { get; set; }
+
+        [Required]
+        public bool IsPublic { get; set; }
```
milana-todorovic: I really like that you have detailed constraint annotations for the database, but I would probably still eventually implement some form of validation either in this class or in the ActionsBenefits service, I'm not sure if you're currently planning to do that at some point? Or if not that, maybe I'd at least put some try catch blocks in the repo and replace the exceptions Entity Framework throws with custom exceptions just to make everything a little less tied to the technology and handle the exceptions a bit easier in the rest of the system.

HealthcareSystem/Backend/Service/ActionBenefitService/RabbitMqActionBenefitService/RabbitMqActionBenefitService.cs
```diff
@@ -0,0 +1,138 @@
...
+    {
+        private IModel _channel;
+        private IConnection _connection;
+        private readonly IServiceProvider _service;
+        private readonly String _hostname;
+        private readonly String _username;
+        private readonly String _password;
+        private readonly String _queueName;
+
+        public RabbitMqActionBenefitMessageingService(IServiceProvider service, IOptions<RabbitMqConfiguration> rabbitMqOPtions)
```
milana-todorovic, line 28: I feel like injecting IPharmacyService and IActionsBenefitsService instead of IServiceProvider to make the dependencies more explicit might be better. 
Also, does the RabbitMQConfiguration need to be added to appsettings.json and read at Startup to be used like this? I don't see a section for it in the appsettings right now, I feel like that will cause issues (like maybe this being null when injected)?

milana-todorovic (commented):
Seems pretty good, I left some comments with suggestions and possible issues. (I'm not approving yet just because of the possible issue with missing RabbitMQ settings. Feel free to decide whatever you like when it comes to the rest of the suggestions since they mostly concern clean code).

HealthcareSystem/Backend/Service/ActionBenefitService/RabbitMqActionBenefitService/RabbitMqActionBenefitService.cs
borislav-puzic: yes it will be later, but that is part of different feature, so ill add it when i do that feature

HealthcareSystem/Backend/Model/Pharmacies/ActionBenefit.cs
borislav-puzic: yes i was planing to do some validation

HealthcareSystem/Backend/Service/ActionBenefitService/RabbitMqActionBenefitService/RabbitMqActionBenefitService.cs
borislav-puzic: Injection must be done like this beacause of different lifetimes of RabbitMqActionBenefitMessageingService(Singleton) and Services(Scope)
For configuration see last commit

HealthcareSystem/Backend/Service/ActionBenefitService/RabbitMqActionBenefitService/RabbitMqActionBenefitService.cs
borislav-puzic: Ok, will make interface. But its not needed for injection, tried it in another project when i was researching rabbitmq and it worked.

milana-todorovic (approved):
Everything seems alright to me!

anciz (approved):

Stefan-Kitanovic (approved):
Everything looks good to me. Code is clean and will make implementation of further features easier.

Stefan-Kitanovic:
Everything looks good to me. Code is clean and should make implementation of further features easier.

borislav-puzic:
[(5) IA3.1 - I as director want to be able to receive action and benefits notifications in order to get better deals [8]](https://trello.com/c/7U9NydO8/35-5-ia31-i-as-director-want-to-be-able-to-receive-action-and-benefits-notifications-in-order-to-get-better-deals-8)

PR #76, opened 2020-11-26
feat[ia5.2]: implementation of api registration form and saving to db

Stefan-Kitanovic:
I have implemented a form to register a pharmacy and added validations for it in frontend and the controller itself. Also the tests cover the post method in the controller.


Stefan-Kitanovic:
[(2) IA5.2 - Form where director can save API key and info to our database](https://trello.com/c/aNaC7zWo/48-2-ia52-form-where-director-can-save-api-key-and-info-to-our-database)

HealthcareSystem/IntegrationAdaptersTests/UnitTests/ApiRegisterTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            };
+            var controller = new HomeController(stubRepository.Object)
+            {
+                TempData = tempData
+            };
+            controller.ModelState.AddModelError("Url", "Required");
+
+            var result = controller.ApiRegister(pharmacy);
+
+            Assert.IsType<BadRequestObjectResult>(result);
```
milana-todorovic: What is being tested in these exactly?  I find it kind of hard to tell from the code.

Stefan-Kitanovic: It's testing if the method in the controller checks correctly if the model passed is valid. That's why I just assert depending on what kind of request it returns. Also TempData stuff is used for the success message in frontend and I have to initialize it here else the object for it never gets created and the test fails.

milana-todorovic: So it essentially tests if your model is annotated correctly, right? Since MVC does the rest of it automatically if it is? If so I suppose that's alright, but I don't really like how much it's tied into MVC (I'm not the biggest fan of how easily the MVC model annotations in the backend can get mixed up with the Entity Framework annotations either, but I suppose there's no easy way to fix that right now).

borislav-puzic (dismissed):
lgtm

HealthcareSystem/IntegrationAdapters/Controllers/PharmacyController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace IntegrationAdapters.Controllers
+{
+    public class PharmacyController : Microsoft.AspNetCore.Mvc.Controller
+    {
+        private readonly PharmacyService _pharmacyService;
+
+        public PharmacyController(IPharmacyRepo iPharmacyRepo)
+        {
+            _pharmacyService = new PharmacyService(iPharmacyRepo);
```
milana-todorovic: There's an IPharmacyService interface, why inject the repo and instantiate the service here instead of just injecting the service? The service is already added as scoped in the startup as well, was this an oversight or were there some issues with DI?

Stefan-Kitanovic: Yeah it was an oversight on my part. Thanks for pointing it out, I just pushed a fix for it

milana-todorovic: Oversight, no serious issues. Fixed in 0dd5b746d447e1725aac9e50581071ff190decad.

HealthcareSystem/IntegrationAdaptersTests/UnitTests/ApiRegisterTests.cs
Stefan-Kitanovic: Yeah basically checks are there in case someone manages to bypass frontend validation and this was the best solution with what we've got right now, could be refactored at a later point if needed

milana-todorovic (commented):
I think I might be having a bug with this, here's the description so you can try to reproduce it or let me know if that's intended behavior for now. The issue happens when the data is put in the form correctly, but the operation fails (which is always for me since my DB connection is down). I assume that this case was meant to be handled, since I'm not seeing an exception in debug so it's clearly caught somewhere. However, I'm not getting an error message in the browser, instead I just get this page:
image
Is this some sort of bug with error handling, or is the error handling just not finished and this case isn't meant to work?

Stefan-Kitanovic:
The page looks like that because it returns an empty object in case of failed validation in the controller, and that's only caused if they bypass frontend validation or like in your case if DB is down. I guess proper error handling in case of DB being down should be added later.

milana-todorovic (approved):
Seems alright.

borislav-puzic (approved):
lgtm

PR #78, opened 2020-11-26
feat[w2.5]: advanced search of examinations and therapies

draganacarapic:
I know that the methods in the services are not clean code, so if you have a suggestion on how to refact these methods, tell me.
* Due to changes in the model, I also committed new migration.


draganacarapic:
[(3) W2.5 As a patient I want to search documentation about my recipes and examination reports with strictly defined conditions so that I could get more precise results.](https://trello.com/c/73CQBIMc/24-3-w25-as-a-patient-i-want-to-search-documentation-about-my-recipes-and-examination-reports-with-strictly-defined-conditions-so-t)

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+
+        /// <summary>
+        /// /getting examinations linked to a certain patient
+        /// </summary>
+        /// <param name="patientJmbg">patients jmbg</param>
+        /// <returns>if alright returns code 200(Ok), if not 404(not found)</returns>
+        [HttpGet("{patientJmbg}/{startDate}/{endDate}/{doctorSurname}/{anamesis}/{firstOperator}/{secondOperator}/{thirdOperator}")]
+        public ActionResult AdvancedSearchExaminations(string patientJmbg, string startDate, string endDate, string doctorSurname, string anamesis, string firstOperator, string secondOperator, string thirdOperator)
```
milana-todorovic: Having this many path parameters is pretty odd, I suggest either using query parameters or a DTO. My personal recommendation would be to use a DTO, so something like this:
```
enum LogicalOperator
{
    And,
    Or
}

class ExaminationSearchDTO 
{
    //fields here
}
```
Also, why are you sending dates as strings, is that because they're path parameters? I think the ability to use DateTime and and enum for logical operators s another reason to use a DTO here, you could possibly get implicit conversions to happen with query parameters but it might be a pain.

HealthcareSystem/PatientWebApp/Controllers/TherapyController.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+
+        /// <summary>
+        /// /getting examinations linked to a certain patient
+        /// </summary>
+        /// <param name="patientJmbg">patients jmbg</param>
+        /// <returns>if alright returns code 200(Ok), if not 404(not found)</returns>
+        [HttpGet("{patientJmbg}/{startDate}/{endDate}/{doctorSurname}/{drugName}/{firstOperator}/{secondOperator}/{thirdOperator}")]
+        public ActionResult AdvancedSearchExaminations(string patientJmbg, string startDate, string endDate, string doctorSurname, string drugName, string firstOperator, string secondOperator, string thirdOperator)
```
milana-todorovic: Same thing here, do a DTO and send actual dates and enum for operators.

HealthcareSystem/PatientWebApp/DTOs/ExaminationDTO.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Threading.Tasks;
+using Model.PerformingExamination;
+
+namespace PatientWebApp.DTOs
+{
+    public class ExaminationDTO
+    {
+        public int Id { get; set; }
+        public string Type { get; set; }
+        public string DateAndTime { get; set; }
```
milana-todorovic, line 13: Now that I'm looking at it, why strings instead of dates in the DTOs as well? I haven't tried in C# / ASP but surely the way DateTime is serialized to JSON has got to be configurable?

HealthcareSystem/PatientWebApp/Mappers/ExaminationMapper.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <summary>
+        /// /addapting object type ExaminationDTO to type Examination 
+        /// </summary>
+        /// <param name="dto">object type ExaminationDTO</param>
+        /// <returns>object type Examination</returns>
+        public static Examination ExaminationDTOToExamination(ExaminationDTO dto)
+        {
+            Examination examination = new Examination();
+            examination.Id = dto.Id;
+            examination.Type = (TypeOfExamination)Enum.Parse(typeof(TypeOfExamination), dto.Type, true);
```
milana-todorovic, line 22: There also definitely has to be a way to get the JSON serializer to parse enums for you automatically / write them nicely as strings instead of numbers? (Unless you really don't want to put enums and dates in DTOs for some reason. In that case also ignore the suggestion above to make the logical operators for advanced search enums and all of this basically.)

HealthcareSystem/Backend/Service/ExaminationAndPatientCard/ExaminationService.cs
```diff
@@ -28,6 +29,336 @@ public void AddExamination(Examination examination)
             _scheduledExaminationRepository.AddExamination(examination);
         }
 
+        public List<Examination> AdvancedSearchExaminationReports(List<Examination> examinations, string startDate, string endDate, string doctorSurname, string anamesis, string firstOperator, string secondOperator, string thirdOperator)
```
milana-todorovic: Ok, so, this is the main thing to fix, since it's obviously not very readable. My suggestion here would be to use a modified version of the [Specification Pattern](https://www.codeproject.com/Articles/670115/Specification-pattern-in-Csharp). The modifications I suggest are to skip the Not specification since you're not using it anyway, to make one LogicalOperation method instead of And and Or to make it easier to use with the exact data you're getting, and to skip the ExpressionSpecification and make specific classes for what you need instead (mostly for purposes of eventual individual testing). 

So, here is approximately how I suggest implementing the advanced search specification for examinations, and how it would be used (with the assumption that the DTO suggestion is implemented as I said, adjust accordingly for changes in that). 

###  Specification
```
    public interface ISpecification<T>
    {
        bool IsSatisfiedBy(T o);
        ISpecification<T> BinaryOperation(LogicalOperator op, ISpecification<T> specification);
    }
```
```
    public abstract class CompositeSpecification<T> : ISpecification<T>
    {
        public ISpecification<T> BinaryOperation(LogicalOperator op, ISpecification<T> specification)
        {
            switch (op)
            {
                case LogicalOperator.And:
                    return new AndSpecification<T>(this, specification);
                case LogicalOperator.Or:
                    return new OrSpecification<T>(this, specification);
                default:
                    return this;
            }
        }

        public abstract bool IsSatisfiedBy(T o);
    }
```
```
    public class AndSpecification<T> : CompositeSpecification<T>
    {
        ISpecification<T> _leftSpecification;
        ISpecification<T> _rightSpecification;

        public AndSpecification(ISpecification<T> left, ISpecification<T> right)
        {
            _leftSpecification = left;
            _rightSpecification = right;
        }

        public override bool IsSatisfiedBy(T o)
        {
            return _leftSpecification.IsSatisfiedBy(o)
                && _rightSpecification.IsSatisfiedBy(o);
        }
    }
```
```
   public class OrSpecification<T> : CompositeSpecification<T>
    {
        ISpecification<T> _leftSpecification;
        ISpecification<T> _rightSpecification;

        public OrSpecification(ISpecification<T> left, ISpecification<T> right)
        {
            _leftSpecification = left;
            _rightSpecification = right;
        }

        public override bool IsSatisfiedBy(T o)
        {
            return _leftSpecification.IsSatisfiedBy(o)
                || _rightSpecification.IsSatisfiedBy(o);
        }
    }
```
```
    public class StartDateSpecification : CompositeSpecification<Examination>
    {
        private DateTime? _startDate;

        public StartDateSpecification(DateTime? startDate)
        {
            _startDate = startDate;
        }

        public override bool IsSatisfiedBy(Examination o)
        {
            if (_startDate is null)
                return true;

            // TODO implement whatever check actually needs to happen here,
            // i'm not sure what start date is meant to be filtering for
            return true;
        }
    }
```
```
    public class EndDateSpecification : CompositeSpecification<Examination>
    {
        private DateTime? _endDate;

        public EndDateSpecification(DateTime? endDate)
        {
            _endDate = endDate;
        }

        public override bool IsSatisfiedBy(Examination o)
        {
            if (_endDate is null)
                return true;

            // TODO implement whatever check actually needs to happen here,
            // i'm not sure what end date is meant to be filtering for
            return true;
        }
    }
```
```
    public class DoctorSurnameSpecification : CompositeSpecification<Examination>
    {
        private string _doctorSurname;

        public DoctorSurnameSpecification(string doctorSurname)
        {
            _doctorSurname = doctorSurname;
        }

        public override bool IsSatisfiedBy(Examination o)
        {
            if (_doctorSurname is null)
                return true;

            return _doctorSurname.Equals(o.Doctor.Surname);
        }
    }
```
```
   public class AnamnesisSpecification : CompositeSpecification<Examination>
    {
        private string _anamnesis;

        public AnamnesisSpecification(string anamnesis)
        {
            _anamnesis = anamnesis;
        }

        public override bool IsSatisfiedBy(Examination o)
        {
            if (_anamnesis is null)
                return true;

            return _anamnesis.Equals(o.Anamnesis);
        }
    }
```
### Usage
```
public class ExaminationSearchDTO
    {
        public DateTime? StartDate { get; set; }
        public LogicalOperator EndDateOperator { get; set; }
        public DateTime? EndDate { get; set; }
        public LogicalOperator AnamnesisOperator { get; set; }
        public string Anamnesis { get; set; }
        public LogicalOperator DoctorSurnameOperator { get; set; }
        public string DoctorSurname { get; set; }
    }
```
```
        public IEnumerable<Examination> AdvancedSearch(string jmbg, ExaminationSearchDTO parameters)
        {
            List<Examination> examinations = repo.GetExaminationsByPatient(jmbg);

            ISpecification<Examination> filter = new StartDateSpecification(parameters.StartDate);
            filter = filter.BinaryOperation(parameters.EndDateOperator, new EndDateSpecification(parameters.EndDate));
            filter = filter.BinaryOperation(parameters.AnamnesisOperator, new AnamnesisSpecification(parameters.Anamnesis));
            filter = filter.BinaryOperation(parameters.DoctorSurnameOperator, new DoctorSurnameSpecification(parameters.DoctorSurname));

            return examinations.Where(examination => filter.IsSatisfiedBy(examination));
        }
```
### Why bother with this?
* Individual filters are easy to test and read. The full specification is also easy to test on an individual examination, and it's instantiation seems relatively readable to me.
* Easy to change in case your search parameters change. What would you do with your current method if you were asked to add a parameter (other than cry, pretty much)? As for what you would do here - maybe add one more specification class, instantiate accordingly, connect to existing parameters any way you like. The backend modification for that would probably take literally a couple minutes. Honestly, even counting the time I spent on adding the boilerplate for logical operators and all that, adding this entire implementation took like half an hour.
### Additional suggestions
* Maybe the object that holds the search parameters could be a bit smarter than a DTO and instantiate the specification object. 
* @PSW-2020-ORG4/devops Any other suggestions to make this better maybe? I wish it was simpler, but this really just isn't a simple functionality. Is Specification overkill here? I feel like any attempt at a simpler implementation I've thought of actually ends up more complicated.

milana-todorovic (commented):
Here are some detailed refactoring suggestions, mostly aimed at making things readable and getting any sort of certainty that this works.

HealthcareSystem/Backend/Service/ExaminationAndPatientCard/ExaminationService.cs
milana-todorovic: Applied in a6a4faba184bae4fbbe467acddb31e9e492749d9.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
milana-todorovic: Applied in a6a4faba184bae4fbbe467acddb31e9e492749d9.

HealthcareSystem/PatientWebApp/Controllers/TherapyController.cs
milana-todorovic: Applied in a6a4faba184bae4fbbe467acddb31e9e492749d9.

HealthcareSystem/PatientWebAppUnitTests/UnitTests/ExaminationControllerTests.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        [Fact]
+        public void Advanced_search_non_exitent_examination()
+        {
+            ExaminationController examinationController = SetupExaminationController();
+
+            var examinationSearchDTOInvalidObject = _objectFactory.GetExaminationSearchDTO().CreateInvalidTestObject();
+            var result = examinationController.AdvanceSearchExaminations(examinationSearchDTOInvalidObject);
+
+            Assert.True(result is NotFoundObjectResult);
```
milana-todorovic, line 84: I feel like the Service should be stubbed for these, not the repository - because for advanced search, the Service doesn't just pass the calls anymore, it has it's own business logic. And this test doesn't seem like it's meant to test that business logic - it seems like it's meant to test whether or not the controller responds with the right messages (assuming that the data it's working with is correct). The messaging logic in the controller itself won't (or at least shouldn't) actually break if the business logic in the service is broken - this test, on the other hand, absolutely will.

HealthcareSystem/PatientWebAppUnitTests/UnitTests/TherapyControllerTests.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        [Fact]
+        public void Advanced_search_non_exitent_therapy()
+        {
+            TherapyController therapyController = SetupTherapyController();
+
+            var therapySearchDTOInvalidObject = _objectFactory.GetTherapySearchDTO().CreateInvalidTestObject();
+            var result = therapyController.AdvanceSearchTherapies(therapySearchDTOInvalidObject);
+
+            Assert.True(result is NotFoundObjectResult);
```
milana-todorovic, line 73: Same thing here, I would stub the Service.

milana-todorovic (commented):
The controller tests will be okay if the Service is stubbed, but I think there are much better things to test here than that. All the test you have right now target the controller and check whether or not it forms messages correctly. That's alright, but the search itself isn't really tested at all. So, here are some tests I propose:

### Test the AdvancedSearch method in the TherapyService and in the ExaminationService
Here's approximately what I think would work nicely here: Stub the repository to contain a list of multiple therapies / examinations. Decide on search parameters and instantiate the DTO. Determine which of the objects from your stub you expect the search to return if it's working correctly. Call the AdvancedSearch method. Check if the contents of the return value are what you expected. The concept of this test is pretty similar to the one shown [here](https://www.youtube.com/watch?v=jCg4DF-srFc&feature=youtu.be&ab_channel=NikolaLuburi%C4%87).

### Test the Specification classes
So, this is a much more lenient suggestion - basically I think these tests would be useful for making sure that your search works 100% correctly, but I do realize that there would be a whole lot of them, and that they might not be super resistant to refactoring (I would maybe ask some of the product owners if these would actually be good tests or not?). The situation I would aim to cover is this - how hard would it be to notice if you did everything here pretty much perfectly, except exactly one individual condition in your combined search? For example, if everything works, except for the condition of when a Therapy matches the start date criteria. 
I propose testing the Specification classes individually to prevent that. So, to test the ExaminationAnamnesisSpecification for example, you'd do something like this:
```
// instantiate the ExaminationAnamnesisSpecification with a parameter of your choice
// instantiate an Examination that's supposed to match that particular search criteria

// call IsSatisfiedBy

// assert that it returned true
```
And similarly for the false case as well. Obviously, the big issue here is that there are many tiny Specification objects, so writing all those tests would be a pain.
I also have an alternative approach, which might actually be nicer and easier - don't do these tiny tests, but do a *lot* of tests on the AdvancedSearch method (and I mean a ***lot***). Make sure to cover the cases where one parameter is given, and the rest are not. Definitely use Theory though, you don't want to be writing the same test over and over again.

HealthcareSystem/PatientWebAppUnitTests/UnitTests/ExaminationControllerTests.cs
```diff
@@ -62,26 +64,173 @@ public void Get_existent_examination_by_non_patient_jmbg()
...
+
+        [Fact]
+        public void Advanced_search_examination_operator_or_anamnesis()
+        {
+            ExaminationService examinationService = new ExaminationService(_stubRepository.CreateExaminationStubRepository());
+            
+            var result = examinationService.AdvancedSearch(new ExaminationSearchDTO(jmbg: "1309998775018", startDate: new DateTime(2020, 10, 10), endOperator: LogicalOperator.Or, endDate: new DateTime(2020, 12, 12), doctorOperator: LogicalOperator.And, doctorSurname: null, anamnesisOperator: LogicalOperator.Or, anamnesis: "lklklkl"));
+
+            Assert.NotEmpty(result);
+        }
```
milana-todorovic: This and the rest of the similar tests are alright, although they could be further improved. Have you considered using Theory, since you have many similar tests here? And I would maybe add more entities in the repo, as far as I can tell there's only one during these tests. Although that definitely would make coming up with the data and correctly determining what will be returned a lot more complicated, I think I don't mind if you leave these as they are.

milana-todorovic (commented):
I generally like the tests you added, but could you split these methods up into a couple more classes? There's just so much in the ExaminationControllerTests for example, I'd extract something like ExaminationAdvancedSearchTests and ExaminationSpecificationTests or something like that from this class. And the same for TherapyControllerTests.

milana-todorovic (dismissed):
Seems good to me!

sladjana-savkovic (dismissed):

milana-todorovic (approved):

Pufke (approved):

PR #82, opened 2020-11-27
Feature g2.0 connection graphical editor graphical editor server backend

kmilena98:
Added graphicalEditoServer to connect graphicalEditor and backend . Added methods for filling database tables Rooms,EquipmentInRooms,ConsumableEquipment and NonConsumableEquipment. Added method for getting equipment for specific room id from database. In mainWindow, you should uncomment lines AddRooms, AddEquipment and AddEquipmentInRoom to fill database with data. You may have problem with connecting graphicalEditor with graphicalEditorServer because of the localhost since we mapped specific localhost from server to functions in ServerService in graphicalEditor


HealthcareSystem/Backend/Controller/RoomAndEquipment/EquipmentController.cs
```diff
@@ -14,8 +14,8 @@ namespace Controller.RoomAndEquipment
 {
    public class EquipmentController
 {
```
Pufke: Fix indents , Ctrl+K, Ctrl+D, which indents the whole file

HealthcareSystem/Backend/Migrations/20201127131042_NMGE.cs
```diff
@@ -0,0 +1,166 @@
+﻿using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Backend.Migrations
+{
+    public partial class NMGE : Migration
```
Pufke, line 6: think out more meaningful name than NMGE

Pufke: A tip for naming things:
1) Describe the thing in a sentence or two (easier to do in Serbian).
2) Remove any words that are not crucial to convey the meaning.
3) Remove any words that can be inferred from the Type, containing function, containing class, or containing module.
4) See if what remains sufficiently describes the thing.
5) Rename.

HealthcareSystem/Backend/Migrations/20201127131152_NMGE2.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Backend.Migrations
+{
+    public partial class NMGE2 : Migration
```
Pufke, line 5: Same problem

HealthcareSystem/Backend/Service/RoomAndEquipment/ConsumableEquipmentService.cs
```diff
@@ -9,33 +9,56 @@
 using Model.Users;
 using Repository;
 using System.Collections.Generic;
+using Backend.Service.RoomAndEquipment;
+using Backend.Repository;
+using Backend.Repository.EquipmentInRoomsRepository;
 
 namespace Service.RoomAndEquipment
```
Pufke, line 13: Fix indents

HealthcareSystem/GraphicalEditor/Service/ServerService.cs
```diff
@@ -0,0 +1,119 @@
+﻿using GraphicalEditor.Enumerations;
+using GraphicalEditor.Models;
+using GraphicalEditor.Models.MapObjectRelated;
+using Model.Manager;
+using RestSharp;
+using System;
+using System.Collections.Generic;
+
+namespace GraphicalEditor.Service
```
Pufke: Methods in this class contains too many lines of code. Generally, any method longer than ten lines should make you start asking questions.

If you feel the need to comment on something inside a method, you should take this code and put it in a new method. Even a single line can and should be split off into a separate method, if it requires explanations. And if the method has a descriptive name, nobody will need to look at the code to see what it does.

To reduce the length of a method body, use Extract Method ( https://refactoring.guru/extract-method ).

Pufke (changes requested):

ogikuzma:
In order to make these functionalities work you will need to follow these steps:
1. Update database with the new migration ( Set GraphicalEditorServer as a startup project and Backend as a default one)
2. Uncomment last 4 lines in MainWindow.xaml.cs (In latest commit 2 lines)
3. Start GraphicalEditorServer
4. Start GraphicalEditor WPF
5. Uncomment the previous 4 lines since they are only used to populate database tables with start data

There is a great chance you will have problems with ports. In order to solve it go to launchSettings.json in the server, copy "applicationUrl" port and paste it into ServerConstants file (GraphicalEditor)

Since AddRooms method is only used the first time you want to fill the database, we have not invested much time into refactoring. Methods for adding other data into tables will be refactored.

HealthcareSystem/GraphicalEditor/Service/ServerService.cs
ogikuzma: Yes, I agree. Refactoring has already started parallel to making a PR.

ogikuzma:
I've just done refactoring.

-  ServerService is separated into RoomService, EquipmentService and InitializeDatabaseData.
- Migration was renamed

ServerService is commented for now. After this PR is approved, my colleague will test it and commit the commented code. Also, take into consideration that adding to the database will be done only once when you first open the project and never again since we have not been requested to enable that functionality, so I believe there isn't need for thorough review for that part.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
```diff
@@ -0,0 +1,51 @@
...
+                _nonConsumableEquipmentService.newNonConsumableEquipment(nonCosumableEquipment);
+            }
+            else
+            {
+                ConsumableEquipment cosumableEquipment = JsonConvert.DeserializeObject<ConsumableEquipment>(final);
+
+                _consumableEquipmentService.newConsumableEquipment(cosumableEquipment);
+            }
+            return Ok();
+        }
```
milana-todorovic: What is going on here? Are you having issues with serialization / deserialization?

milana-todorovic (commented):
Things seem generally okay as far as establishing an initial connection is concerned, which I assume was your main goal here? 

One thing I don't really like is that you have a lot of Controllers and Services in the Backend part for Equipment, ConsumableEquipment, NonConsumableEquipment, EquipmentInRooms, and so on, and it's really hard to say what each of their responsibilities is and really hard to read them. I feel like that's unnecessarily complicated right now and might warrant a redesign.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
ogikuzma: - The goal was to establish a connection with our WPF and database
- The problem with different classes in Backend is that we were bounded with the existing SiMS project which we did not want to change in order not to break something that is used by the web team. I agree, it can make things more complicated.
- As serialization/deserialization is concerned, we were having some weird problems with it, so this is the only way we made it to work

milana-todorovic: What sort of problems? Maybe we can help, this is really odd and shouldn't be happening. Although, I see that the GraphicalEditor references Backend, are you using the Model classes directly? I really don't think that should be like that. Separating with a DTO layer will probably make the serialization issues go away too.
Also, I don't think Web is using equipment at all right now, you can probably do whatever you want there. 
Merging the initial project with the established connection to develop and then changing it on a different branch might be a better approach for all of this right now though.

VasilijeBursac:
[(3) g2.0: Create and connect database for storing rooms, medicine and equipment informations](https://trello.com/c/vCuVaCd8/40-3-g20-create-and-connect-database-for-storing-rooms-medicine-and-equipment-informations)

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
ogikuzma: We tried serialization of the object we want to send and deserialize in controllers, but it was constantly giving us exceptions. Maybe, it is worth trying again. Since that approach did not work we converted an object into json string, but these brackets {} could not be added, so we added them into the controller with [FromBody] annotation and did deserialization. It was pretty weird, so I believe it is worth trying again.
Yes, we are using the Model classes for now. Adding DTOs is in the plan, but we are waiting to see what will the next require from us, so if needed, refactoring will be done. Take into consideration that addition of the objects into the database will be done only once. Update and delete will be done through the server only.
We will need to consult with Web before we start making any changes. I will tell our SM to investigate that part on the next meeting.

milana-todorovic: It's the hierarchy probably, the serializer can't tell what the actual type of the object is. Having DTOs without the hierarchy (or maybe even removing it from the model, I don't think it actually adds much) is probably the way to go in the future.

ogikuzma: I am planning to give a new try soon, so I will see if there something I missed. I don't really think it is the hierarchy, I have already done serialization of objects in the hierarchy and it worked fine.

milana-todorovic: No, I'm pretty sure it is the hierarchy. Look at what this method is doing - checking the length of the message and then instantiating different types based on that. I assume that the type of the parameter that couldn't get deserialized was the base of the hierarchy? JSON wouldn't know what type to instantiate then, it can't really deal with polymorphism out of the box.
Splitting this method into two might make these issues go away though, and it would make sense anyway.

ogikuzma: If we are talking only about this method and checking the number of parameters, that is correct. That is the reason why it was made like that in the first place. I will split this method into 2. Since adding new items is not required from us ( for now, and there is a chance it will stay like that in the future, but I cannot tell for sure until new sprints come) I believe investing any more time on furthermore refactoring of add functionality would be a waste of time since we are limited with the deadline. After we consult with the Web team, we will maybe do redesign of the model and addition of DTO, so that serialization/deserialization becomes smoother.

milana-todorovic: But I see the methods are still receiving strings, it still won't deserialize these? That's really odd then. 
But yeah, I agree with not refactoring anything here, and probably not wasting any more time on whatever is wrong with the serialization either. I'm fine with merging this branch whenever the conflict is solved.

HealthcareSystem/Backend/Controller/RoomAndEquipment/EquipmentController.cs
milana-todorovic: Solved in 227dd73d685d299759a4b4d3dbf857be00385e68 as far as I can see.

HealthcareSystem/Backend/Service/RoomAndEquipment/ConsumableEquipmentService.cs
milana-todorovic: Solved in 227dd73d685d299759a4b4d3dbf857be00385e68.

HealthcareSystem/GraphicalEditor/Service/ServerService.cs
milana-todorovic: Solved by splitting up the class in 227dd73d685d299759a4b4d3dbf857be00385e68 as far as I can see.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
ogikuzma: I've tried a few different options for transferring data, but this is the only one that seems to work. I am open to refactoring as soon as I see a chance to do it smoothly.
Could you take a look at the message I left on Discord regarding the conflict?

Pufke (approved):

Pufke (dismissed):

milana-todorovic (dismissed):
Everything seems alright to me. I think the merge didn't cause issues but @PSW-2020-ORG4/devops someone should still check if everything builds and runs.

milana-todorovic (dismissed):
Seems alright.

anciz (dismissed):

ogikuzma:
It seems like adding the medicine to this PR will probably take longer than expected, so I will merge this branch as soon as this conflict resolve is approved.

milana-todorovic (dismissed):
Seems ok, build works.

milana-todorovic (approved):

anciz (approved):

PR #84, opened 2020-11-27
feat[w2.3]:html and js for patient profile

jelenabudisa:
This pull request is for patient viewing his profile.


jelenabudisa:
[(3) W2.3 As a patient I want to observe my profile information so that I could check if my data is correct.](https://trello.com/c/CbUobywv/22-3-w23-as-a-patient-i-want-to-observe-my-profile-information-so-that-i-could-check-if-my-data-is-correct)

anciz (approved):
Seems to be good.

milana-todorovic (approved):
Seems okay.

PR #91, opened 2020-11-29
refactor[all]:handling exceptions

sladjana-savkovic:
- I have a goal to eliminate potential errors such unexpected exceptions.


draganacarapic (dismissed):

milana-todorovic:
Can you elaborate your goal for this PR? Basically, error handling is one of the things I like the least in your project and I have a lot of suggestions and opinions. Is the goal here to substantially improve it, or just to patch up a few specific situations so they don't cause server failure? And if it's the second, do you want me to leave comments on the issues I see with your error handling in general here so that you can apply it for the whole application sometime in the future, or should I assume that you already know and intend to fix things at some point?

sladjana-savkovic:
> Can you elaborate your goal for this PR? Basically, error handling is one of the things I like the least in your project and I have a lot of suggestions and opinions. Is the goal here to substantially improve it, or just to patch up a few specific situations so they don't cause server failure? And if it's the second, do you want me to leave comments on the issues I see with your error handling in general here so that you can apply it for the whole application sometime in the future, or should I assume that you already know and intend to fix things at some point?

The goal is to patch up a few specific situations about exceptions, which could be a risk for this sprint.

milana-todorovic:
> The goal is to patch up a few specific situations about exceptions, which could be a risk for this sprint.

Alright, then it will be approved, but do you want detailed comments on what to improve in the future?

sladjana-savkovic (commented):
I change url in ajax call about filling survey for patient, because last url was a mistake.

draganacarapic (approved):

HealthcareSystem/Backend/Service/CityAndCountry/CityService.cs
```diff
@@ -18,7 +18,14 @@ public CityService(ICityRepository cityRepository)
...
         {
-            return _cityRepository.GetCities();
+            try
+            {
+                return _cityRepository.GetCities();
+            }
+            catch (Exception)
+            {
+                throw new NotFoundException("There is no cities in database.");
+            }
```
milana-todorovic, line 28: It's good that you're preventing server failure here, but it's really bad that the message you will propagate to the rest of your server and your client is completely incorrect. You will never get an exception from the database for fetching an empty table - you'll just get an empty list. So here are some things that an exception here could actually mean: the database connection is down, the database schema is out of sync with the class model so this table doesn't exist or conversion fails, and so on. How you want to handle such situations and how detailed the error responses will be in your application is your decision of course, but at the very least the message that reaches your client in this situation should be something like 500 Internal Server Error, and definitely not 404 Not Found. 
Somewhat related, I think that returning 404 for empty queries instead of 200 with an empty list like you're doing right now is a valid choice, but I feel like you're really inconsistent when it comes to where that's handled. My instinct here is that only controllers should be concerned with this particular choice, so I would have the service layers return empty lists with no exceptions thrown and just check for that in the controller.

HealthcareSystem/Backend/Service/ExaminationAndPatientCard/PatientCardService.cs
```diff
@@ -22,7 +22,14 @@ public PatientCardService(IActivePatientCardRepository activePatientCardReposito
...
         {
-            _activePatientCardRepository.AddPatientCard(patientCard);
+            try
+            {
+                _activePatientCardRepository.AddPatientCard(patientCard);
+            }
+            catch (Exception)
+            {
+                throw new DatabaseException("Patient card with id=" + patientCard.Id + " already exists in database.");
+            }
```
milana-todorovic, line 32: Again, this is not necessarily what is happening here, this error message might be completely wrong. Basically, when handling database exceptions I'd at the very least have one catch that handles exceptions for constraint violations (so that would end up propagating 400 Bad Request to the client) and one catch that handles other things (which the client should probably see as 500 Internal Server Error or something like that).

milana-todorovic (approved):
Approved as the goal here is to prevent server failure, not provide proper error handling. Some comments about general error handling are provided for future consideration.

Basically, here is the big issue with error handling I've seen so far in your application: you are losing information while preventing server failure, and as a consequence of that actively propagating wrong information about what actually happened to your client. And that's not just bad from a design standpoint for your hypothetical future users - at a very practical and currently relevant level I'm sure it makes debugging a complete nightmare for you.

The first step for fixing this would be to just put more thought into it at the team level, decide on a convention, and apply it consistently. I actually think one of your first steps as a team when you started making a REST server should've been deciding what status codes you want to use and exactly when (and in my personal opinion, the minimum set should be something like 200 OK, 201 Created, 400 Bad Request, 401 Forbidden, 403 Unauthorized, 404 Not Found, 500 Internal Server Error), and then deciding how to actually propagate information about some of these situations to the controllers that actually form the response (which would include designing an internal system of exceptions with precise meanings and such). Obviously, this discussion would've eventually included how to handle issues detected during validation at application level, issues with validation detected at database level, various internal issues such as database failures and so on. And I feel like this discussion either didn't happen or wasn't given enough consideration, and that as a consequence of that you all just do whatever you think might work on a case by case basis, which means that things end up pretty inconsistent and that you end up wasting effort on figuring out things that should be convention decided by the team.

Also, somewhat unrelated but I'm putting it here because I also keep seeing issues with this and I think the problem has a similar root - what is your approach to general design of your REST API (including resource naming, HTTP method selection, and so on)? Because I think that is also something that should be done as a team, preferably right at the start of sprints when you see what endpoints you will need to add.

PR #92, opened 2020-11-29
feat[g2.3]: adding new projects for graphicalEditor and graphicalEdit…

kmilena98:
Adding new projects so we can implement tests for every feature in graphicalEditor and graphicalEditorServer


milana-todorovic (approved):
Seems alright, projects are in the solution and build works.

anciz (approved):

PR #96, opened 2020-12-01
feat[ia7.2]: sftp services, frontend and drug consumption report

Stefan-Kitanovic:
This branch contains implementations that were done over the weekend for the sprint review, which includes a report about drug consumption in given date range, implementation of SFTP service and frontend for it.


Stefan-Kitanovic:
[(3) IA7.2 - I as a director want to periodically send reports about medical consumption to pharmacies in order to receive suitable actions and benefits. [2]](https://trello.com/c/gG5qU7mw/43-3-ia72-i-as-a-director-want-to-periodically-send-reports-about-medical-consumption-to-pharmacies-in-order-to-receive-suitable-ac)

Stefan-Kitanovic:
[(3) IA7.1 - I as a director want to be able to create a report of medication consumption in order to determine priorities in procurement of medications. [2]](https://trello.com/c/1eZ0NOtf/36-3-ia71-i-as-a-director-want-to-be-able-to-create-a-report-of-medication-consumption-in-order-to-determine-priorities-in-procurem)

HealthcareSystem/Backend/Repository/DrugConsumptionRepository/MySqlDrugConsumptionRepository/MySqlDrugConsumptionRepository.cs
```diff
@@ -0,0 +1,72 @@
...
+                .Join(
+                    _context.Drugs,
+                    dc => dc.DrugId,
+                    d => d.Id,
+                    (dc, d) => new
+                    {
+                        DrugId = d.Id,
+                        DrugName = d.Name,
+                        TotalConsumption = dc.Quantity
+                    })
```
milana-todorovic, line 58: I'm not the most familiar with Entity Framework, but I believe it might be possible to avoid manually writing a join here if you add a reference to Drug in consumption along with the id? It would probably do an implicit join via the lazy loading proxy when you accessed that field then. I'm not 100% what EF will do implicitly and what not though, I definitely don't mind how you implemented things here. I really like that you're doing this nicely with the group by.

HealthcareSystem/Backend/Service/DrugConsumptionService/DrugConsumptionService.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public void UpdateDrugConsumption(DrugConsumption dc)
+        {
+            _drugConsumptionRepository.UpdateDrugConsumption(dc);
+        }
+
+        public void DeleteDrugConsumption(DrugConsumption dc)
+        {
+            _drugConsumptionRepository.DeleteDrugConsumption(dc);
+        }
```
milana-todorovic: (Read the comment bellow first so things make more sense, I added them in the wrong order.)

Is there really a use case that requires these to be modifiable? I feel like they should be historical records - you lower the quantity of a drug, and make a DrugConsumption to record that it happened. It's not something I see ever being updated or deleted (except maybe by DB admins after a certain amount of time when we no longer need the history). 
Even if such a use case exists though, I definitely not your current one, so I probably wouldn't have added these methods either.

HealthcareSystem/Backend/Service/DrugConsumptionService/DrugConsumptionService.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public DrugConsumption GetDrugConsumptionById(int id)
+        {
+            return _drugConsumptionRepository.GetDrugConsumptionById(id);
+        }
+
+        public void CreateDrugConsumption(DrugConsumption dc)
+        {
+            _drugConsumptionRepository.CreateDrugConsumption(dc);
```
milana-todorovic: I don't think this should just be created on it's own like this. The Drug class has the quantity field in it - so really, the way I understand it, the saved DrugConsumption instances should be a historical record of how that field changed. So I feel like the more correct business logic would be something like this - you have a service that deals with the hospital's drug storage in general. It has methods to reduce and increase the current quantity of a Drug, which can also create a DrugConsumption as a side effect. And then it would also make sense to me to put getters for things like current quantity of a drug and DrugConsumptions and DrugConsumptionRecords there.
I don't think this use case required you to actually implement adding DrugConsumptions yet, you really just need the getters to make reports? So I probably would've preferred just adding the methods you need at this point to the Service layer.

HealthcareSystem/Backend/Service/SftpService/ISftpService.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Renci.SshNet;
+using Renci.SshNet.Sftp;
+
+namespace Backend.Service.SftpService
+{
+    public interface ISftpService
```
milana-todorovic: I think the professor said while talking to the Graphical Editor team yesterday that a better name for things like these might be Communicator rather than Service? And that naming something Service implies that it's responsibility is orchestration of domain objects, which isn't really true here. I don't mind leaving the names as is right now though and refactoring it later, since we have other classes like this one which are currently named Service and when they're renamed some package refactoring should happen as well.

HealthcareSystem/Backend/Service/SftpService/SftpService.cs
```diff
@@ -0,0 +1,97 @@
...
+            using var client = new SftpClient(_config.Host, _config.Port == 0 ? 22 : _config.Port, _config.Username, _config.Password);
+            try
+            {
+                client.Connect();
+                return client.ListDirectory(remoteDirectory);
+            }
+            catch (Exception exception)
+            {
+                Console.WriteLine(exception);
+                return null;
```
milana-todorovic: Will this have proper error handling eventually? At least a message about failure reaching the client, since I don't think recovery is exactly possible here.

HealthcareSystem/IntegrationAdapters/Controllers/SftpController.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Backend.Service.DrugConsumptionService;
+using Backend.Service.SftpService;
+using Microsoft.AspNetCore.Mvc;
+using Newtonsoft.Json;
+using System;
+using Backend.Model.Pharmacies;
+
+namespace IntegrationAdapters.Controllers
+{
+    public class SftpController : Microsoft.AspNetCore.Mvc.Controller
```
milana-todorovic, line 9: So the use case here is just sending a consumption report to a pharmacy, right? I feel like there should be a bit more abstraction happening here, I don't like the fact that we're doing something via SFTP reaching all the way to the controller since it's just an implementation detail. Think something like, IDrugConsumptionReportService (bad name probably, but you know your use cases better than me and can probably name it better). And then a specific implementation of that Service that happens to use the SftpService (really, SftpCommunicator) to send out the reports.

milana-todorovic (commented):
Here are some design suggestions mostly. I would like if you considered them and discussed them further, but I don't require you to apply any of this if you disagree.

HealthcareSystem/Backend/Model/Pharmacies/SftpConfig.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Threading.Tasks;
+
+namespace Backend.Model.Pharmacies
+{
+    public class SftpConfig
+    {
+        public string Host { get; set; } = "192.168.1.3";
+        public int Port { get; set; } = 22;
+        public string Username { get; set; } = "tester";
+        public string Password { get; set; } = "password";
```
milana-todorovic: Pull these values out into appconfig.json and load them in Startup (same as RabbitMQ settings are loaded in your project, or mail settings in the PatientWebApp).

milana-todorovic (changes requested):
This will be an issue for deployment so it needs to be changed.

HealthcareSystem/Backend/Model/Pharmacies/SftpConfig.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Backend.Model.Pharmacies
+{
+    public class SftpConfig
+    {
+        public string Host { get; set; } = "192.168.1.3";
```
Pufke: Hardcoding IP addresses is security-sensitive. It has led in the past to the following vulnerabilities:

 * CVE-2006-5901(http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2006-5901)
 * CVE-2005-3725(http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2005-3725)

Today's services have an ever-changing architecture due to their scaling and redundancy needs. It is a mistake to think that a service will always have the same IP address. When it does change, the hardcoded IP will have to be modified too. This will have an impact on the product development, delivery and deployment:

The developers will have to do a rapid fix every time this happens, instead of having an operation team change a configuration file.
It forces the same address to be used in every environment (dev, sys, qa, prod).
Last but not least it has an effect on application security. Attackers might be able to decompile the code and thereby discover a potentially sensitive address. They can perform a Denial of Service attack on the service at this address or spoof the IP address. Such an attack is always possible, but in the case of a hardcoded IP address the fix will be much slower, which will increase an attack's impact.

Pufke: image

HealthcareSystem/Backend/Model/Pharmacies/SftpConfig.cs
Stefan-Kitanovic: I completely forgot to get rid of these after I finished setting up SFTP. Just pushed a fix for it.

HealthcareSystem/IntegrationAdapters/Controllers/SftpController.cs
Stefan-Kitanovic: Report sending was just the first use of SFTP that was implemented. As far as I know, this sprint involves implementations of other use cases requiring SFTP, so I think the controller is fine as it is right now, especially considering we are still not sure how the frontend will look in the end.

HealthcareSystem/Backend/Service/SftpService/SftpService.cs
Stefan-Kitanovic: Yeah I think when we completely figure out what we want to do with frontend we should do a PR just for error handling like this and have them reported to the user, as this is not the only place that's missing error handling.

HealthcareSystem/Backend/Service/SftpService/ISftpService.cs
Stefan-Kitanovic: I actually wrote these classes before our session with the professor, but yeah I agree they should be renamed. Perhaps have a separate folder called Communicators or something and move all classes such as this one there? I could rename it right now or I could refactor it later. I don't mind either.

HealthcareSystem/Backend/Service/DrugConsumptionService/DrugConsumptionService.cs
Stefan-Kitanovic: I agree that we don't need methods such as this right now. Having DrugConsumptions table be updated from other services such as general Drug service or some other services that deal with our stock seems like a much better idea. For now I have removed these unnecessary CRUD operations as they are not needed for this feature. At some point later if someone is working on a feature that involves changing drug quantity in the storage, they could also just tie in DrugConsumptions.

HealthcareSystem/Backend/Repository/DrugConsumptionRepository/MySqlDrugConsumptionRepository/MySqlDrugConsumptionRepository.cs
Stefan-Kitanovic: Honestly, I am not that well versed in EF either. I actually started by writing an SQL query for what I needed and then turned it into Linq syntax. If someone who's more experienced with EF knows how to make this better, they are more than welcome to, but I think it's fine as it is for now.

HealthcareSystem/Backend/Model/Pharmacies/SftpConfig.cs
milana-todorovic: Fixed in a6b683487e40f42a1f66bbea3d1bd1d87c6d26f7.

HealthcareSystem/Backend/Model/Pharmacies/SftpConfig.cs
milana-todorovic: Fixed in a6b683487e40f42a1f66bbea3d1bd1d87c6d26f7.

HealthcareSystem/Backend/Service/DrugConsumptionService/DrugConsumptionService.cs
milana-todorovic: Fixed in 6498babbb3b08e9824f980dab9090775e0555339.

HealthcareSystem/Backend/Service/DrugConsumptionService/DrugConsumptionService.cs
milana-todorovic: Fixed in 6498babbb3b08e9824f980dab9090775e0555339.

HealthcareSystem/Backend/Service/SftpService/ISftpService.cs
milana-todorovic: Yeah, that's what I was thinking pretty much. Maybe we could do that for these classes now if you don't mind, and change the older classes to match later?

Stefan-Kitanovic: Sure I don't mind. Perhaps we could do it during the company meeting soon just so everyone is on the same page when it comes to naming conventions.

HealthcareSystem/IntegrationAdapters/Controllers/SftpController.cs
milana-todorovic: Okay, I'm alright with leaving it as is for now then, although the fact that SFTP will have other uses here actually reinforces the point that it shouldn't be used by the controllers directly. So, as I understand, the exact use case here is approximately this: user chooses the dates and hits the send report button, the report is generated, the generated report is then sent to the pharmacy (or all pharmacies?). 
So here is a really long and rambling discussion of the layer structure for this:
- The Controller has a method that takes the dates and maybe like the pharmacy id if that's necessary. It passes all of that to the corresponding Service method, and just deals with things like catching exceptions, converting error messages, and such (although I'm not sure how any of that goes with MVC and how necessary it is, I'm much better acquainted with REST Controllers). In REST such a Controller would also be easily testable because you could just make a bunch of stubs of the Service that generate appropriate exceptions, and then check if the return messages you're getting are correct. This is probably not the case with MVC though, since there's redirection to the View happening and such?
- The Service also has a method that directly corresponds to the use case, so it also takes the dates and possibly pharmacy id. The Service then orchestrates between the other parts of the system, so here it would do something like call the Repository to get the data for the report based on the dates, possibly also call another Repository to get data needed for communication with the specific pharmacy by the id (or maybe communication data for all pharmacies if that's what we're doing instead), and then call the Communication layer to actually send the message.
- So, the Communication layer, or whatever it ends up being called. I would actually make that have a domain level abstraction at the very top as well, not just the generic classes for specific protocols. The reason for that is really just to prevent changes to the Service layers when the protocol changes. Actually, here's a thing I just realized, isn't this particular domain object or set of domain objects exactly what the goal of the project here is though? I feel like what I'm talking about here would actually just be the famous IPharmacyAdapter interface or something? Anyway, the point is that for this use case this particular class would have a method that's really just something like SendReports which takes a list of the DrugConsuptionReport objects which the Service fetched, and also gets the data needed for communication somehow (either the Service also fetched that and injects it here, or it was already in this class somehow, I really haven't worked what would be good for that and what wouldn't but I think it'll get clearer after we learn DDD). This way the Service isn't dealing with forming a specific message from the report list or anything, and it's nicely abstracted away from any communication protocols. As a consequence, the Service or Service and Repository together or whatever are also super easy to test, cause you can just mock the IPharmacyAdapter interface and check if the correct method gets called with the correct parameters.  And the method in a particular implementation of IPharmacyAdapter or whatever it's called would then form a protocol appropriate message and delegate sending it to the generic ISftpCommunicator or whatever. Now that I think about it, this is actually kind of similar to the discussion about loggers and domain loggers in particular what we needed to read for last weeks questionnaire? Any kind of message sending ends up being similar in design I suppose. Also, this all works out quite nicely with SFTP or HTTP or something, but applying it to RabbitMQ and hiding away the whole background service might be a whole other can of worms that I'd have no idea where to start with.

Anyway, obviously many things still need to be worked out better here, but I hope it gives you some useful ideas.

Stefan-Kitanovic: I appreciate the input and ideas. I think this will require a lot of designing as the whole team, but for now I would leave it like this so everyone can start working on their features and later on when we properly design everything I will just refactor it.

milana-todorovic (approved):
Looks good, you'll have to merge develop for status checks to pass though.

HealthcareSystem/IntegrationAdapters/Controllers/SftpController.cs
milana-todorovic: Yeah, I agree, I would also leave it as is for now, I just thought it was a good starter for a discussion.

HealthcareSystem/Backend/Service/SftpService/ISftpService.cs
milana-todorovic: Applied in 83fc2347f54ff5e0e58491584f824e21d6768c1d.

Stefan-Kitanovic:
Just merged it. Thanks for all the feedback!

anciz (approved):

PR #99, opened 2020-12-02
refactor[all]: refactoring backend code

draganacarapic:
* I made a new class for a doctor's specialty (model, repository, service).
* I added two new fields to the class Examination.


draganacarapic:

anciz (dismissed):

HealthcareSystem/Backend/Model/PerformingExamination/Examination.cs
```diff
@@ -13,32 +13,35 @@
...
+
         [ForeignKey("Room")]
         public int IdRoom { get; set; }
         public virtual Room Room { get; set; }
 
         [ForeignKey("PatientCard")]
         public int IdPatientCard { get; set; }
+        public bool IsCanceled { get; set; }
+        public bool IsSurveyCompleted { get; set; }
         public virtual PatientCard PatientCard { get; set; }
```
milana-todorovic: Is having an IsCanceled boolean enough for your use cases? Are you sure something like an ExaminationStatus enumeration with Scheduled, Cancelled, and Finished (where it will switch to finished after the doctor saves the report through his app) wouldn't be simpler to use in some cases? (To clarify, I'm not asking you to do it the way I said, I'm asking you to consider and maybe discuss with your team what fits your needs better.)

Also, kind of petty, but could you move these fields around? Putting these between the PatientCard and it's id is really ugly, move IsCancelled and IsSurveyCompleted up to the fields that aren't foreign keys (right bellow Anamnesis).

HealthcareSystem/Backend/Model/Users/Specialty.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+using Model.Users;
+
+namespace Backend.Model.Users
+{
+    public class Specialty
+    {
+        [Key]
+        public int Id { get; set; }
```
milana-todorovic: Is this missing the DatabaseGenerated annotation?

HealthcareSystem/Backend/Service/UsersAndWorkingTime/SpecialtyService.cs
```diff
@@ -0,0 +1,29 @@
...
+        public List<Specialty> GetSpecialties()
+        {
+            try
+            {
+                return _specialtyRepository.GetSpecialties();
+            }
+            catch (Exception)
+            {
+                throw new NotFoundException("There is no specialties in database.");
+            }
```
milana-todorovic: Since we agreed that I would point out and give a detailed explanation when I see issues like these with error handling, I see two issues here. First, I would definitely move this try catch to the Repository, not the Service, because exceptions here will be thrown by Entity Framework and it's the Repository layer's job to abstract and hide the Entity Framework (or whatever data access) and the exceptions from it shouldn't reach the Service.
Second, and worse, this error message is plain wrong. A database will *never* throw an exception when fetching an empty table. An exception here could mean a lot of things, such as the database is down, the table doesn't exist at all, and so on, but never that no entities were found. If you want to throw not found if there are no specialties, I think you should just check in the Controller if the list you got from the Service is empty, and then return an appropriate HTTP code.

HealthcareSystem/Backend/Model/Users/Doctor.cs
```diff
@@ -10,7 +11,10 @@ namespace Model.Users
    public class Doctor : User
    {
         public string NumberOfLicence { get; set; }
-        public TypeOfDoctor Type { get; set; }
+
+        [ForeignKey("Specialty")]
+        public int SpecialtyId { get; set; }
+        public virtual Specialty Specialty { get; set; }
```
milana-todorovic: This isn't too relevant, but wouldn't making this a list of specialties make more sense? It doesn't make using it more complicated and it might be more accurate for real life, doctors aren't limited to having just one specialty.

HealthcareSystem/Backend/Model/PerformingExamination/Examination.cs
milana-todorovic: Applied in f48613f08e758ec1d13738d35d386e2834e035de.

HealthcareSystem/Backend/Service/UsersAndWorkingTime/SpecialtyService.cs
milana-todorovic: Fixed in f48613f08e758ec1d13738d35d386e2834e035de.

HealthcareSystem/Backend/Model/Users/Doctor.cs
milana-todorovic: Applied in f48613f08e758ec1d13738d35d386e2834e035de.

HealthcareSystem/Backend/Model/Users/DoctorSpecialty.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Backend.Model.Users
+{
+    public class DoctorSpecialty
+    {
+        public string DoctorJmbg { get; set; }
+        public virtual Doctor Doctor { get; set; }
+        public int SpecialtyId { get; set; }
+        public virtual Specialty Specialty { get; set; }
+    }
```
milana-todorovic, line 14: I don't think you actually need to have this class in the object model. Look at [this](https://www.entityframeworktutorial.net/code-first/configure-many-to-many-relationship-in-code-first.aspx), Entity Framework should be capable of generating a table for the relationship on it's own. So it should be enough to just have a virtual list of Specialty objects in Doctor, and a virtual list of Doctor objects in Specialty. I feel like that would be more convenient to use.

milana-todorovic: Correction: this is apparently a pretty new feature in EF Core 5.0, and not available in EF Core 3.1 which we are using. The way it's currently implemented is the best (and possibly only) way to do it in 3.1.

milana-todorovic (dismissed):

anciz (dismissed):

anciz (approved):

milana-todorovic (approved):

PR #102, opened 2020-12-02
feat[g2.3]:merging nonConsumable and Consumable Equipment into Equipment

kmilena98:
Merging nonConsumable and ConsumableEquipment into Equipment and addition it into database. We are planning to create DTO-s for communication between GraphicalEditor and GraphicalEditorServer and to move some functions from Service from GraphicalEditor to GraphicalEditorServer so we can delete Model from GraphicalEditor and use DTO-s instead.


HealthcareSystem/Backend/Service/RoomAndEquipment/EquipmentService.cs
```diff
@@ -0,0 +1,61 @@
...
+            _equipmentInRoomRepository = equipmentInRoomRepository;
+        }
+        public Equipment newEquipment(Equipment equipment)
+        {
+            return _equipmentRepository.NewEquipment(equipment);
+        }
+        public List<Equipment> ViewEquipment()
+        {
+            return _equipmentRepository.GetAllEquipment();
+        }
```
milana-todorovic: Could you rename these methods so they match the naming pattern from the rest of the Services? I believe we're currently using Add and Get for naming these methods.

HealthcareSystem/Backend/Model/Manager/TypeOfEquipment.cs
```diff
@@ -1,14 +1,24 @@
...
-        NON_CONSUMABLE = 1
+        NEEDLE = 0,
+        MASK = 1,
+        BEND = 2,
+        DESINFECTANT = 3,
+        BED = 4,
+        TABLE = 5,
+        CHAIR = 6,
+        COMPUTER = 7,
+        INSTRUMENT = 8
```
milana-todorovic, line 21: I think you should turn this into a class, like what was done with Specialty. It's fields would just be Id and Name, really, but it would be useful because it would allow us to add as many equipment types as we like instead of being limited by this enumeration. And you could also add other fields if you want, maybe something like Description or IsConsumable (it might be good to still keep some distinction between a table and a needle).

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
```diff
@@ -16,35 +16,20 @@ namespace GraphicalEditorServer.Controllers
...
-            ConsumableEquipment cosumableEquipment = JsonConvert.DeserializeObject<ConsumableEquipment>(JSONContent);
-            _consumableEquipmentService.newConsumableEquipment(cosumableEquipment);
-
-            return Ok();
-        }
-
-        [HttpPost("nonconsumable")]
-        public ActionResult AddNonConsumableEquipment([FromBody] String JSONString)
+        [HttpPost]
+        public ActionResult AddEquipment([FromBody] String JSONString)
```
milana-todorovic: Is there a reason why this is still like this? If you just left it as is and haven't tried switching to just having Equipment as the parameter, I'd think you should do that. If you have tried and it didn't work I'd like to see the error message you got.

HealthcareSystem/Backend/Model/Manager/Equipment.cs
```diff
@@ -1,17 +1,35 @@
...
+ * Purpose: Definition of the Class ConsumableEquipment
  ***********************************************************************/
 
 using System;
-
+using System.ComponentModel.DataAnnotations.Schema;
 
 namespace Model.Manager
 {
     public class Equipment
```
milana-todorovic, line 13: I like that you got rid of the hierarchy, it wasn't really adding any value to the model I think.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
ogikuzma: We are planning to add DTOs probably after adding GUI changes as a separate PR, so that is when we will try refactoring methods similar to this one.

HealthcareSystem/Backend/Model/Manager/TypeOfEquipment.cs
ogikuzma: Are you sure this would be the right approach for this problem since when writing the name of an equipment object you could easily make a typo and instead of bend get a bed for example? I cannot find the Specialty class, is it merged into develop?

milana-todorovic: Yeah, I think Specialty is on develop, merged just yesterday. And I'm not exactly sure when the issue with typos could happen - do you mean when selecting an equipment type? And the answer to that is that with decent HCI it would never be typed in those cases, you'd still have the user select one of the existing equipment types from a combo box or something. The only difference is that instead of having a fixed enumeration tell you what those existing types are, you would query the database to find out.
And if you mean that typos could happen while administrating the equipment types and adding new ones (which would be done either by the database admin, or exposed to a user with admin privileges like the manager for example) - yes, of course typos could happen, but I fail to see an issue with that. Whoever is administrating would be able to see the typo and just fix it whenever. 
But yeah, generally speaking, it is 100% the right approach pretty much whenever you have a type-instance sort of relationship. Enumerations are fine when you have something with values that won't change unless really extreme circumstances happen, and that also doesn't have that many values, like state of an object or something. But for something like this, where it's absolutely reasonable that types can be added and removed at will, an enumeration just doesn't cut it. 
Of course, you don't have to implement this if you don't have the time or don't like it as a design choice (although it's objectively a better design choice than what we currently have).
(Edit: I just now see that's it's actually been implemented in 3e5d3d3aa45304258414aaf176e9eaa88533e8ca)

HealthcareSystem/Backend/Model/Manager/EquipmentType.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace Backend.Model.Manager
+{
+    public class EquipmentType
+    {
+        [Key,DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public int Id { get; set; }
+        public string EquipementType { get; set; }
```
milana-todorovic: I think something like Name might be a better name for this.

HealthcareSystem/Backend/Repository/EquipmentTypeRepository/IEquipmentTypeRepository.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Backend.Repository
+{
+   public interface IEquipmentTypeRepository
+    {
+        EquipmentType GetEquipmentType(int id);
+        List<EquipmentType> GetAllEquipmentTypes();
+        EquipmentType SetEquipmentType(EquipmentType equipmentType);
```
milana-todorovic: I think we generally used Update for this, not Set?

HealthcareSystem/Backend/Service/RoomAndEquipment/IEquipmentTypeService.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+
+namespace Backend.Service.RoomAndEquipment
+{
+    public interface IEquipmentTypeService
+    {
+        EquipmentType AddEquipmentType(EquipmentType equipmentType);
+        EquipmentType GetEquipmentTypeById(int id);
+        List<EquipmentType> GetEquipmentTypes();
+        EquipmentType EditEquipmentType(EquipmentType equipmentType);
```
milana-todorovic: Why is it Edit here, and Set in the repository? I think Update was generally used for both.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentTypeController.cs
```diff
@@ -0,0 +1,51 @@
...
+            List<EquipmentType> equipmentTypes = new List<EquipmentType>();
+            try
+            {
+                equipmentTypes = _equipmentTypeService.GetEquipmentTypes();
+                return Ok(equipmentTypes);
+            }
+            catch (NotFoundException exception)
+            {
+                return NotFound(exception.Message);
+            }
```
milana-todorovic, line 43: What would throw a NotFound exception here?

milana-todorovic (dismissed):
Seems alright.
I left you a couple comments about naming / consistency to possibly fix since you'll be making another commit to fix the conflict anyway.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentTypeController.cs
kmilena98: I am not sure if this method will ever throw exception, but if it does it should mean that there were exception while getting allEquipmentTypes, l guess nothing more precisely then that. I made that method based on other get methods that we previously done. Do you think l should change something there?

HealthcareSystem/Backend/Model/Manager/TypeOfEquipment.cs
ogikuzma: I wanted to recommit yesterday, but Milena has already done it.
I believe enums could make some use cases smoother, but of course, this approach is always available and could be really useful for a wide range of cases. Since Milena has already made changes, I believe we could resolve this part of the conversation.

milana-todorovic (approved):
Seems alright.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentTypeController.cs
milana-todorovic: Yeah, so, this is something I've been discussing with the Web team quite a bit recently, since they had quite a few issues with their error handling (check out PR #91 and PR #99 if you're interested). Basically, in this particular case, NotFoundException is one of our internal exceptions, correct? You only need to catch that if you explicitly throw it somewhere as a message for your higher layers and want to catch it in the Controller for example to deal with that message, but it's not going to come out of nowhere. So since I don't think you throw a NotFoundException anywhere in your Service and Repository, this is never going to catch anything. The <code>_equipmentTypeService.GetEquipmentTypes();</code> call specifically *could* throw exceptions, but they'll actually be exceptions from various Entity Framework failures since you don't have any try catch blocks in the Repository to catch those and either handle them somehow, or replace them with your own exceptions with precise meanings.
But yeah, if what you're going for here is to return 404 when no equipment types are found, a more accurate way to do that here would just be to check if the list you are getting is empty, and return 404 if it is.
I'm not really insisting you change anything here though, error handling is kind of a large topic that I don't think you've really done anything with in GraphicalEditorServer so far. If you want to tackle that you should probably discuss with your team (and possibly the Web team as well) exactly what your goals are with it and how you want to proceed.

ogikuzma (approved):

kmilena98:
NonConsumable and ConsumableEquipment are merged into class Equipment and communication between GraphicalEditor and GraphicalEditorServer is improved.

HealthcareSystem/Backend/Service/RoomAndEquipment/IEquipmentTypeService.cs
milana-todorovic: Fixed in 52728b8266cccd1bd69b043a7e6060b01c8ed960

HealthcareSystem/Backend/Repository/EquipmentTypeRepository/IEquipmentTypeRepository.cs
milana-todorovic: Fixed in 52728b8266cccd1bd69b043a7e6060b01c8ed960

HealthcareSystem/Backend/Model/Manager/EquipmentType.cs
milana-todorovic: Fixed in 52728b8266cccd1bd69b043a7e6060b01c8ed960

HealthcareSystem/Backend/Service/RoomAndEquipment/EquipmentService.cs
milana-todorovic: Fixed in 3e5d3d3aa45304258414aaf176e9eaa88533e8ca

VasilijeBursac:
[(3) g2.3: Overview of equipment and medicine: 5h[3]](https://trello.com/c/AQBpJl1G/36-3-g23-overview-of-equipment-and-medicine-5h3)

PR #103, opened 2020-12-03
feat[w3.8]: survey after examination

jelenazeljko55:
Tasks:
add validation;
add some methods in the controller and service;
modify survey js file;
modify test for survey.


milana-todorovic:
You changed the constructor of the SurveyController without changing how the constructor is called in tests, so now compilation of the test project fails. Please fix before review.

jelenazeljko55:
[(3) W3.8 : As a patient I want to have the opportunity to fill out a survey for a completed examination, so I can share my experience quickly.](https://trello.com/c/YJ19qu11/35-3-w38-as-a-patient-i-want-to-have-the-opportunity-to-fill-out-a-survey-for-a-completed-examination-so-i-can-share-my-experience)

HealthcareSystem/PatientWebApp/Controllers/SurveyController.cs
```diff
@@ -20,29 +21,39 @@ public class SurveyController : ControllerBase
...
         public ActionResult AddSurvey(SurveyDTO surveyDTO)
         {
             try
             {
                 _surveyValidator.ValidateSurveyFields(surveyDTO);
+                _examinationValidator.CheckIfExaminationIsFinished(surveyDTO.ExaminationId);
+                _examinationValidator.CheckIfSurveyAboutExaminationIsCompleted(surveyDTO.ExaminationId);
                 _surveyService.AddSurvey(SurveyMapper.SurveyDTOToSurvey(surveyDTO));
+                _examinationService.CompleteSurveyAboutExamination(surveyDTO.ExaminationId);
+                return Ok();
```
milana-todorovic: I'm not a fan of the fact that basically, the Controller is doing orchestration here, not the Services. I'm also not a fan of the fact that you're fetching the Examination from the Repository three separate times, twice by the Validator and once by the Service. Basically, here's approximately what I think should be happening - this is all done by the SurveyService instead. And the method in SurveyService would do something like this - fetch the Examination, call a method on the Examination to mark that a survey has been completed for it (this method in Examination would check if the examination is finished and that a survey isn't already completed before setting the field), instantiate the Survey objects from the DTO (possibly with some validation happening in the constructor) and add them to the database, update the Examination in the database.
However, my idea of what should be happening is fairly tentative right now, I would probably lave this as is until we go over DDD and then refactor.

One thing that I probably would like to be fixed right now on the other hand - you're not handling what happens if no Examination with this id exists. CheckIfExaminationIsFinished, CheckIfSurveyAboutExaminationIsCompleted, and CompleteSurveyAboutExamination will all end up throwing NullPointerExceptions if the Repository returns null, and it's not replaced by a local exception or handled anywhere.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -165,5 +165,22 @@ public ActionResult GetFollowingExaminationsByPatient(string patientJmbg)
                 return StatusCode(500, e.Message);
             }
         }
+
+
+        [HttpGet("examination-by-id/{examinationId}")]
+        public ActionResult GetExaminationById(int examinationId)
```
milana-todorovic: This isn't in accordance with [RESTful resource naming](https://restfulapi.net/resource-naming/), it should be just "examinations/{id}" and not "examinations/examination-by-id/{id}".

HealthcareSystem/PatientWebApp/Controllers/SurveyController.cs
```diff
@@ -45,15 +45,19 @@ public ActionResult AddSurvey(SurveyDTO surveyDTO)
                 _surveyService.AddSurvey(SurveyMapper.SurveyDTOToSurvey(surveyDTO));
                 _examinationService.CompleteSurveyAboutExamination(surveyDTO.ExaminationId);
                 return Ok();
-            }
-            catch (ValidationException exception)
+            }         
+            catch (NotFoundException exception)
             {
-                return BadRequest(exception.Message);
+                return NotFound(exception.Message);
```
milana-todorovic, line 51: I would probably still return BadRequest here, not NotFound. NotFound generally only makes sense to me when the user is explicitly trying to fetch a resource which doesn't exist, which isn't the case here.

milana-todorovic (approved):
Seems okay to me.

anciz (approved):

Pufke (approved):

PR #104, opened 2020-12-03
feat[w3.5]: cancel appointment

draganacarapic:
* I will commit the tests on this pull request when I pull from develop the refactored code about the tests, which is being worked on by a colleague from the team.


milana-todorovic:
I renamed your PR so it's obvious what functionality it refers to.

draganacarapic:

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -65,5 +68,29 @@ public ActionResult AdvanceSearchExaminations(ExaminationSearchDTO examinationSe
...
+            {
+                return BadRequest(exception.Message);
+            }
+            catch (DatabaseException exception)
+            {
+                return StatusCode(500, exception.Message);
+            }
+            _examinationService.CancelExamination(id);
+            return Ok();
+        }
```
milana-todorovic: So, one quick thing to fix - the Service method call can also throw exceptions. When writing Controllers, I would generally put the entire happy path in the try catch, so this would be something like:
```
            try
            {
                _examinationValidator.CheckIfExaminationCanBeCanceled(id);   
                _examinationService.CancelExamination(id);
                return Ok();
            }
            catch (ValidationException exception)
            {
                return BadRequest(exception.Message);
            }
            catch (DatabaseException exception)
            {
                return StatusCode(500, exception.Message);
            }
```
Another thing that kind of bothers me is that you fetch the Examination from the database twice, once from the Validator, and once from the Service. Basically, here's how I would fix that - get rid of the Validator in this case entirely. Instead, make a Cancel method in Examination. Have that method check whether the state transition can happen or not, and throw an exception if it can't. Actually, I feel like that's approximately what we'll be asked to do with pretty much all validations for DDD anyway - so because of that I also don't mind if you keep it as is for now and refactor later.

HealthcareSystem/PatientWebAppTests/CreateObjectsForTests/StubRepository.cs
```diff
@@ -68,10 +68,14 @@ public IExaminationRepository CreateExaminationStubRepository()
         {
             var examinationStubRepository = new Mock<IExaminationRepository>();
 
+            Examination examinationValid = _objectFactory.GetExamination().CreateValidTestObject();
+            Examination examinationInvalid = _objectFactory.GetExamination().CreateInvalidTestObject();
```
milana-todorovic: So, here's a thing that bothers me - these objects are actually both valid, it's just that one is cancellable and one is not. I would rename these methods to GetCancellableTestObject and GetUncancellableTestObject, or something like GetCreatedTestObject or GetCancelledTestObject, just so you know what they do later.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
milana-todorovic: Fixed the Controller method in 8d297a7a201cb557715f22f218845f03c8793464. Moving the Validation postponed until next sprint, since we'll start applying DDD then.

HealthcareSystem/PatientWebAppTests/CreateObjectsForTests/StubRepository.cs
milana-todorovic: Fixed in 8d297a7a201cb557715f22f218845f03c8793464.

milana-todorovic (approved):
Seems alright.

anciz (approved):

PR #107, opened 2020-12-04
bugfix-ia3-rabbitmq-fixes

borislav-puzic:
Currentli subscribe unsubscribe for actions and benefits is only applied after app restart
I want to make it posible to do that while application is running


milana-todorovic:
The code for connection retrying breaks your tests for some reason, I'd just remove it for now. I think we can first figure out if we actually need this to launch with Docker Compose and then try to see what's up with the tests. Thanks for trying it out though!

Pufke:
four tests are not performed correctly

image

milana-todorovic (approved):
Seems alright.

anciz (approved):
Alright!

PR #108, opened 2020-12-04
Feature g2.0 create and connect database for storing medicine informations

djura023:
-Implemented classes for storing medicine informations in database.
-It is finished.


VasilijeBursac:
[(3) g2.0: Create and connect database for storing rooms, medicine and equipment informations: 20h [8]](https://trello.com/c/vCuVaCd8/40-3-g20-create-and-connect-database-for-storing-rooms-medicine-and-equipment-informations-20h-8)

HealthcareSystem/Backend/Repository/DrugInRoomRepository/IDrugInRoomRepository.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+   public interface IDrugInRoomRepository
+    {
+        List<DrugInRoom> GetDrugsInRoomByRoomNumber(int roomNumber);
+        DrugInRoom SetDrug(DrugInRoom drugInRoom);
+        bool DeleteDrug(int id);
+        DrugInRoom AddDrug(DrugInRoom drugInRoom);
+        DrugInRoom GetDrugInRoomByDrugId(int drugId);
+    }
+}
```
vlaksi: Why we use DrugInRoom as a return type instead of the void (I think on AddDrug & SetDrug methods)? And why is return type bool for DeleteDrug? I think we should use the void return type and use exceptions.

djura023: There is posibility that we will change just added objects although this disrupts the cqs.

vlaksi: Return type DrugInRoom is okay but if you have time you could change it to void. But try to remove return type bool in DeleteDrug and make void return type & exceptions handling.

vlaksi (approved):
Seems fine.

Pufke (approved):

PR #110, opened 2020-12-05
feat[w3.5]: add restrict for examination cancellation

draganacarapic:
* I have added a ban on canceling an examination if that examination is in the next 48 hours
* This validation is in the Validator layer like all other checks.


vlaksi (approved):
Seems fine.

Pufke (approved):

draganacarapic:

PR #112, opened 2020-12-05
feat[w3.7]: blocking the patient

jelenazeljko55:
I added a method to block the patient

(admin can change patient status)

### DIFF
--- a/HealthcareSystem/Backend/Model/Users/Patient.cs
+++ b/HealthcareSystem/Backend/Model/Users/Patient.cs
@@ -16,19 +16,19 @@ namespace Model.Users
     {
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public DateTime DateOfRegistration { get; set; }
-        public bool IsGuest { get; set; }
+        public bool IsBlocked { get; set; }
         public bool IsActive { get; set; }
         public string ImageName { get; set; }
         public virtual ICollection<Feedback> Feedbacks { get; set; }
         public virtual PatientCard PatientCard { get; set; }
         public Patient()
         {
-            IsGuest = false;
+            IsBlocked = false;
         }
 
         public Patient(string jmbg,string name,string surname,DateTime dateOfBirth,GenderType gender,City city,string homeAddress,
                         string phone,string email,string username,string password,
-                        DateTime dateOfRegistration,bool isGuest,string imagePath)
+                        DateTime dateOfRegistration,bool isBlocked, string imagePath)
         {
             Jmbg = jmbg;
             Name = name;
@@ -51,10 +51,11 @@ namespace Model.Users
             Username = username;
             Password = password;
             DateOfRegistration = dateOfRegistration;
-            IsGuest = isGuest;
+            IsBlocked = isBlocked;
             IsActive = false;
             ImageName = imagePath;
         }
+
         public Patient(Patient patient)
         {
             Jmbg = patient.Jmbg;
@@ -78,7 +79,7 @@ namespace Model.Users
             Username = patient.Username;
             Password = patient.Password;
             DateOfRegistration = patient.DateOfRegistration;
-            IsGuest = patient.IsGuest;
+            IsBlocked = patient.IsBlocked;
             IsActive = false;
             ImageName = patient.ImageName;
         }
--- a/HealthcareSystem/Backend/Repository/PatientRepository/MySqlPatientRepository/MySqlActivePatientRepository.cs
+++ b/HealthcareSystem/Backend/Repository/PatientRepository/MySqlPatientRepository/MySqlActivePatientRepository.cs
@@ -44,7 +44,17 @@ namespace Repository
         }
         public Patient GetPatientByJmbg(string jmbg)
         {
-            return _context.Patients.Find(jmbg);
+            try
+            {
+                Patient patient = _context.Patients.Find(jmbg);
+                if (patient == null)
+                    throw new DatabaseException("Patient doesn't exist in database.");
+                return _context.Patients.Find(jmbg);
+            }
+            catch (Exception)
+            {
+                throw new DatabaseException("The database connection is down.");
+            }
         }
         public void UpdatePatient(Patient patient)
         {
--- a/HealthcareSystem/Backend/Repository/PatientRepository/MySqlPatientRepository/MySqlActivePatientRepository.cs
+++ b/HealthcareSystem/Backend/Repository/PatientRepository/MySqlPatientRepository/MySqlActivePatientRepository.cs
@@ -42,20 +42,23 @@ namespace Repository
         {
             return _context.Patients.ToList();
         }
+
         public Patient GetPatientByJmbg(string jmbg)
         {
+            Patient patient;
             try
             {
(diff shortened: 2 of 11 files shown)


jelenazeljko55:
[(3) W3.7 : As an admin I want to block malicious patients, so that I could prevent bad hospital time managing.](https://trello.com/c/ezzyzi8U/33-3-w37-as-an-admin-i-want-to-block-malicious-patients-so-that-i-could-prevent-bad-hospital-time-managing)

HealthcareSystem/PatientWebApp/wwwroot/js/malicious_patients.js
```diff
@@ -18,13 +18,41 @@
...
+}
+
+
+function blockPatient(patientJmbg) {
+
+    $.ajax({
+        type: "PUT",
+        url: "/api/patient/blocked/" + patientJmbg,
+        success: function () {
+            alert("You have successfully blocked the patient.");
```
Pufke: Here are some reasons why you should not use alerts in your code:

As you said, a lot alert boxes may annoy the users
Styling of alert boxes is as far as I know impossible. Furthermore every browser displays them diffrent, which may not look good in all cases with your website-design
Most modern / common browsers block alerts after a site tried to display a few of them, to prevent alert spamming
You can use jQuery UI dialogs as an alternative, or anything else that is able to display modal dialogs. That way you can customize / style the dialog box and browsers don't block them.

I generally use alert boxes only for debugging / error reporting, not more, but always remember alerts stop the execution of the code for a purpose. Otherwise, you should be using console.log because alert can actually change the result of your code if your code involves asynchronous logic, so you should make sure that it won't affect any asynchronous logic that you may have running.

HealthcareSystem/Backend/Model/Users/Patient.cs
```diff
@@ -16,19 +16,19 @@ public class Patient : User
     {
         [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
         public DateTime DateOfRegistration { get; set; }
-        public bool IsGuest { get; set; }
```
milana-todorovic, line 19: Why remove this? Are guest accounts no longer a thing in our system or?

HealthcareSystem/Backend/Repository/PatientRepository/MySqlPatientRepository/MySqlActivePatientRepository.cs
```diff
@@ -44,7 +44,17 @@ public List<Patient> GetAllPatients()
...
+            {
+                Patient patient = _context.Patients.Find(jmbg);
+                if (patient == null)
+                    throw new DatabaseException("Patient doesn't exist in database.");
+                return _context.Patients.Find(jmbg);
+            }
+            catch (Exception)
+            {
+                throw new DatabaseException("The database connection is down.");
+            }
```
milana-todorovic: This is something @sladjana-savkovic noticed yesterday, but this implementation has some issues. Specifically, you throw an exception at line 51, but you're inside a try block, and then catch Exception. So the DatabaseException you throw here will get caught and replaced anyway (and I believe it should be a NotFoundException you're throwing instead of DatabaseException, too?).

HealthcareSystem/Backend/Service/UsersAndWorkingTime/PatientService.cs
```diff
@@ -86,5 +86,17 @@ public List<Patient> ViewMaliciousPatients()
             }
             return result;
         }
+
+        public void BlockPatient(string jmbg)
+        {
+            Patient patient = GetPatientByJmbg(jmbg);
```
milana-todorovic, line 92: So this is more of a taste thing, but I'd maybe catch the NotFoundException the Repository could throw here and replace it with BadRequest?

HealthcareSystem/Backend/Model/Users/Patient.cs
jelenazeljko55: the guest no longer exists in the system

Pufke (dismissed):

HealthcareSystem/Backend/Repository/PatientRepository/MySqlPatientRepository/MySqlActivePatientRepository.cs
milana-todorovic: Resolved in f220bf17acde3b181f97f446b33ccf25ff9c819a.

HealthcareSystem/Backend/Repository/PatientRepository/MySqlPatientRepository/MySqlActivePatientRepository.cs
```diff
@@ -42,10 +42,23 @@ public List<Patient> GetAllPatients()
...
+            {
+                patient = _context.Patients.Find(jmbg);
+            }
+            catch (Exception)
+            {
+                throw new DatabaseException("The database connection is down.");
+            }
+            if (patient == null)
+                throw new NotFoundException("Patient doesn't exist in database.");
+            return patient;
```
milana-todorovic, line 59: This is better, but your Controller method isn't catching it, if this actually happens you'll get server failure. Add another catch block for NotFoundException in the method.

HealthcareSystem/PatientWebApp/Controllers/PatientController.cs
```diff
@@ -175,5 +175,20 @@ public IActionResult GetMaliciousPatients()
             }
         }
 
+
+        [HttpPut("blocked/{jmbg}")]
+        public ActionResult CancelExamination(string jmbg)
```
milana-todorovic: Also, I just noticed now, why is this called CancelExamination? I assume it's an accident, change to BlockPatient or something.

HealthcareSystem/Backend/Repository/PatientRepository/MySqlPatientRepository/MySqlActivePatientRepository.cs
milana-todorovic: Fixed in be98915f313550c2f0c40f3f02babbc905a76b36

HealthcareSystem/PatientWebApp/Controllers/PatientController.cs
milana-todorovic: Fixed in be98915f313550c2f0c40f3f02babbc905a76b36.

HealthcareSystem/PatientWebApp/Controllers/PatientController.cs
```diff
@@ -177,13 +177,17 @@ public IActionResult GetMaliciousPatients()
...
+        public ActionResult BlockPatient(string jmbg)
         {
             try
             {
                 _patientService.BlockPatient(jmbg);
                 return Ok();
             }
+            catch (NotFoundException exception)
+            {
+                return NotFound(exception.Message);
```
milana-todorovic, line 189: I suppose a discussion could be had on whether NotFound or BadRequest is a more appropriate return code, but either is good enough in this particular case I think.

milana-todorovic (approved):
Seems fine!

sladjana-savkovic (approved):

PR #113, opened 2020-12-05
feat[w3.1]:basic search appointment backend

sladjana-savkovic:
I have the following goals:

- Implement basic search logic
- Add method in controller and test implemented logic in Postman

### DIFF
--- a/HealthcareSystem/Backend/Model/DTO/BasicAppointmentSearchDTO.cs
+++ b/HealthcareSystem/Backend/Model/DTO/BasicAppointmentSearchDTO.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Backend.Model.Exceptions;
+using System;
 using System.Collections.Generic;
 
 namespace Backend.Model.DTO
@@ -10,5 +11,15 @@ namespace Backend.Model.DTO
         public ICollection<int> RequiredEquipmentTypes { get; set; }
         public DateTime EarliestDateTime { get; set; }
         public DateTime LatestDateTime { get; set; }
+
+        public void IsAppointmentValid()
+        {
+            if (string.IsNullOrEmpty(DoctorJmbg))
+                throw new ValidationException("Doctor jmbg cannot be null or empty.");
+            if (EarliestDateTime == null)
+                throw new ValidationException("Earliest date cannot be null.");
+            if (LatestDateTime == null)
+                throw new ValidationException("Latest date cannot be null");
+        }
     }
 }
--- a/HealthcareSystem/Backend/Service/ExaminationAndPatientCard/FreeAppointmentSearchService.cs
+++ b/HealthcareSystem/Backend/Service/ExaminationAndPatientCard/FreeAppointmentSearchService.cs
@@ -1,7 +1,8 @@
 ﻿using Backend.Model.DTO;
-using Backend.Repository;
 using Backend.Repository.ExaminationRepository;
 using Backend.Repository.RoomRepository;
+using Model.Enums;
+using Model.Manager;
 using Model.PerformingExamination;
 using System;
 using System.Collections.Generic;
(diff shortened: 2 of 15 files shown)


sladjana-savkovic:
[(8) W3.1 : As a patient I want to schedule new appointment, so that I could avoid waiting in hospital.](https://trello.com/c/jLGaFLeV/28-8-w31-as-a-patient-i-want-to-schedule-new-appointment-so-that-i-could-avoid-waiting-in-hospital)

Pufke (approved):

HealthcareSystem/Backend/Model/DTO/BasicAppointmentSearchDTO.cs
```diff
@@ -10,5 +11,15 @@ public class BasicAppointmentSearchDTO
...
         public DateTime LatestDateTime { get; set; }
+
+        public void IsAppointmentValid()
+        {
+            if (string.IsNullOrEmpty(DoctorJmbg))
+                throw new ValidationException("Doctor jmbg cannot be null or empty.");
+            if (EarliestDateTime == null)
+                throw new ValidationException("Earliest date cannot be null.");
+            if (LatestDateTime == null)
+                throw new ValidationException("Latest date cannot be null");
```
milana-todorovic, line 22: DateTime isn't nullable, I don't think this is necessary. Actually, now that I look at it, I think we could survive without having this method at all, since we also already check if the doctor exists in the database and get false there if the Jmbg is null or empty.

vlaksi (approved):
Seems fine.

HealthcareSystem/Backend/Service/ExaminationAndPatientCard/FreeAppointmentSearchService.cs
```diff
@@ -42,14 +64,61 @@ public ICollection<Examination> SearchWithPriorities(AppointmentSearchWithPriori
...
+
         private ICollection<DateTime> GenerateStartTimes(DateTime earliest, DateTime latest)
         {
-            throw new NotImplementedException();
+            ICollection<DateTime> startTimes = new List<DateTime>();
+
+            for(DateTime time = earliest; DateTime.Compare(time, latest) < 0; time = time.Add(_appointmentDuration))
+                startTimes.Add(time);
+   
+            return startTimes;
```
milana-todorovic, line 79: One small issue I see with this, you trust the earliest date you're passed in too much. You start generating from earliest without modifying it's time component, so this will break if the time component of earliest isn't 'whole' from the perspective of the appointment duration. You also never check if it's actually in the future (or, within our time limit, it was that everything needs to be scheduled 48 hours in advance, right?).
I would do something like this to get the first time instead of just using earliest:
```
//check if earliest is within the time limit, replace with earliest according to time limit if not

//transform time component
TimeSpan time = earliest.TimeOfDay;
int timeTicks = time.Ticks;
int slotTicks = _appointmentDuration.Ticks;
if (timeTicks % slotTicks != 0)
{
    timeTicks = timeTicks + slotTicks - timeTicks % slotTicks;
    TimeSpan newTime = TimeSpan(timeTicks);
    earlest = earliest.Date + newTime;
}
```
Note: view this as pseudo code, I haven't actually tried compiling and running it. And it would probably make sense for the replacement in case of breached time limit to already have a valid time, so the second part would only happen if earliest was already within the time limit.
Also, it would probably make sense to encapsulate the relatively low level time logic into a separate class sooner or later, but I can't really think of a nice way to do that right now.

milana-todorovic (commented):
I generally really like how this turned out, I would just add the bit of time smoothing I suggested bellow to make sure all the appointments we generate have nice and round start times (and to avoid breaking the search if someone passes in that they want their appointment to start at 12:08 or something).

PR #115, opened 2020-12-05
hotfix: fix DrugService instantiation issue on PatientWebApp startup

draganacarapic:
A dependency was added to the DrugService class in #108, but the Startup class in the PatientWebApp wasn't updated accordingly. This lead to a runtime exception because the DI container couldn't instantiate the DrugService.


milana-todorovic (approved):

milana-todorovic:
I changed the title and description of the PR to explain what the issue was.

sladjana-savkovic (approved):

PR #116, opened 2020-12-06
Feature g2.3 testing equipment

kmilena98:
Testing getting equipment by room number with unit tests


HealthcareSystem/GraphicalEditorServerTests/DataFactory/ICreateTestObjectFactory.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Text;
+
+namespace GraphicalEditorServerTests.DataFactory
+{
+    public interface ICreateTestObjectFactory<T>
+    {
+        T CreateValidTestObject();
+        T CreateValidTestObject(int id);
+        T CreateInvalidTestObject();
+    }
```
milana-todorovic: We had something like this interface in the PatientWebApp tests, but in the end we decided it was too restrictive and wasn't bringing any value, so we got rid of it. I don't really see a need to introduce it to your project as well.

HealthcareSystem/Backend/Service/RoomAndEquipment/EquipmentService.cs
```diff
@@ -68,7 +68,10 @@ public Equipment GetEquipmentById(int equipmentId)
         public List<Equipment> GetEquipmentByRoomNumber(int roomNumber)
         {
             List<EquipmentInRooms> equipmentInRooms = _equipmentInRoomRepository.GetEquipmentInRoomsByRoomNumber(roomNumber);
-
+            if (equipmentInRooms == null) 
+            {
+                return null;
+            }
```
milana-todorovic: Will this ever be null? Why would the Repository return null instead of an empty list? I don't think this should be happening.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
```diff
@@ -42,8 +42,17 @@ public ActionResult AddEquipment([FromBody] Equipment equipment)
...
         public IActionResult GetEquipmentByRoomNumber(int roomNumber)
         {
-            List<Equipment> equipmentsInRoom = _equipmentService.GetEquipmentByRoomNumber(roomNumber);
-            return Ok(equipmentsInRoom);
+            try
+            {
+                List<Equipment> equipmentsInRoom = _equipmentService.GetEquipmentByRoomNumber(roomNumber);
+                if (equipmentsInRoom == null) {
+                    return NotFound("NotFound");
+                }
```
milana-todorovic: Again, why null? Empty list makes more sense and it'd be easier to work with.

HealthcareSystem/GraphicalEditorServerTests/DataFactory/CreateEquipmentType.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace GraphicalEditorServerTests.DataFactory
+{
+    public class CreateEquipmentType : ICreateTestObjectFactory<EquipmentType>
+    {
+        public EquipmentType CreateInvalidTestObject()
+        {
+            return new EquipmentType("fsdafafas",false);
```
milana-todorovic: Why is this not valid?

HealthcareSystem/GraphicalEditorServerTests/DataFactory/CreateEquipmentType.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public EquipmentType CreateValidTestObject()
+        {
+            return new EquipmentType("table",true);
+        }
+
+        public EquipmentType CreateValidTestObject(int id)
+        {
+            return new EquipmentType("table", true);
+        }
```
milana-todorovic: You can see that the interface is too restrictive here already, it makes no sense to send an int here, you don't need it.

HealthcareSystem/GraphicalEditorServerTests/DataFactory/CreateEquipment.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class CreateEquipment : ICreateTestObjectFactory<Equipment>
+    {
+        public Equipment CreateInvalidTestObject()
+        {
+            return new Equipment(-1,null);
+        }
+        public Equipment CreateValidTestObject(int equipmentQuantity)
+        {
+            return new Equipment(equipmentQuantity, new Backend.Model.Manager.EquipmentType("table",true));
```
milana-todorovic: Is there a reason to use EquipmentType like this instead of a using directive?

HealthcareSystem/GraphicalEditorServerTests/DataFactory/ICreateTestObjectFactory.cs
kmilena98: Alright, l have removed that part in last commit

HealthcareSystem/Backend/Service/RoomAndEquipment/EquipmentService.cs
kmilena98: Yes, you are right about that, l made mistake, so that part of code actually returned null, but it should not do that.
I have changed it in last commit too.

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
kmilena98: I have changed that part

HealthcareSystem/GraphicalEditorServerTests/DataFactory/CreateEquipmentType.cs
kmilena98: I thought that name is not valid, or doesnt mean anything, but l changed that too

milana-todorovic: That does make sense semantically, it's just something we couldn't possibly actually validate.

HealthcareSystem/GraphicalEditorServerTests/DataFactory/CreateEquipmentType.cs
```diff
@@ -5,21 +5,16 @@
...
 namespace GraphicalEditorServerTests.DataFactory
 {
-    public class CreateEquipmentType : ICreateTestObjectFactory<EquipmentType>
+    public class CreateEquipmentType 
     {
         public EquipmentType CreateInvalidTestObject()
         {
-            return new EquipmentType("fsdafafas",false);
+            return null;
         }
```
milana-todorovic: I'd probably delete this method entirely instead of returning null.

milana-todorovic (dismissed):
Seems alright.

HealthcareSystem/GraphicalEditorServerTests/DataFactory/CreateEquipmentType.cs
kmilena98: Alright, thanks for advice

VasilijeBursac:
[(3) g2.3: Overview of equipment and medicine: 5h[3]](https://trello.com/c/AQBpJl1G/36-3-g23-overview-of-equipment-and-medicine-5h3)

VasilijeBursac (approved):
Looks good to me.

milana-todorovic (approved):
Seems alright to me.

PR #118, opened 2020-12-06
feat[w3.1]:add examination backend

sladjana-savkovic:
- Modify conditions in appointment service (our team has agreed that an examination can be scheduled at the earliest
tomorrow and that the working time of the hospital is from 7 to 17h)
- Add method for scheduling examination in controller and optionally change DTO, adapter and validation layers


anciz (dismissed):
Seems to be fine.

anciz (approved):
Good

draganacarapic (approved):

sladjana-savkovic:
[(8) W3.1 : As a patient I want to schedule new appointment, so that I could avoid waiting in hospital.](https://trello.com/c/jLGaFLeV/28-8-w31-as-a-patient-i-want-to-schedule-new-appointment-so-that-i-could-avoid-waiting-in-hospital)

PR #120, opened 2020-12-08
feat[g2.1]: Search for objects on the map

VasilijeBursac:
This branch contains code for searching map objects. 

- Using static attribute _allMapObjects from MainWindow definitely isn't the best option but this will be fixed and a special class will be created at the end of this sprint probably or maybe during next week.

### DIFF
--- a/HealthcareSystemFrontends/GraphicalEditor/Controllers/MapObjectController/MapObjectController.cs
+++ b/HealthcareSystemFrontends/GraphicalEditor/Controllers/MapObjectController/MapObjectController.cs
@@ -1,4 +1,5 @@
 ﻿using GraphicalEditor.Models;
+using GraphicalEditor.Models.MapObjectRelated;
 using GraphicalEditor.Repository;
 using GraphicalEditor.Services;
 using GraphicalEditor.Services.Interface;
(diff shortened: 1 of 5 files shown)


VasilijeBursac:
[(2) g2.1: Search for objects on the map: 4h [2]](https://trello.com/c/zdABDaG8/34-2-g21-search-for-objects-on-the-map-4h-2)

milana-todorovic (approved):
I definitely agree with you about using the static field not being a good idea, but since you'll be refactoring anyway I think this is okay enough for now. One thing I'd just say is that maybe things would've been easier for you if you'd just left the SearchMapObjects method logic in the MainWindow for now anyway since the field it needs is there, it might be a little harder to untangle this when refactoring now.

VasilijeBursac:
> I definitely agree with you about using the static field not being a good idea, but since you'll be refactoring anyway I think this is okay enough for now. One thing I'd just say is that maybe things would've been easier for you if you'd just left the SearchMapObjects method logic in the MainWindow for now anyway since the field it needs is there, it might be a little harder to untangle this when refactoring now.

Thanks for the advice! I agree, but it would increase the amount of code in MainWindow so it is easier for us to leave it like this for now...

kmilena98 (approved):
Looks good to me.

PR #121, opened 2020-12-08
gui[g-general]: Implemented Back Button in Side Panel GUI

VasilijeBursac:
On this branch I have implemented Back Button inside Side Panel GUI, so now the user can go one step back during his navigation through the app.

### DIFF
--- a/HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml.cs
+++ b/HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml.cs
@@ -77,13 +77,23 @@ namespace GraphicalEditor
             }
         }
 
-        private Stack<int> _menuSelectionHistory = new Stack<int>();
+        private int? _previousSelectedMenuOptionIndex;
+        public int? PreviousSelectedMenuOptionIndex
+        {
+            get { return _previousSelectedMenuOptionIndex; }
+            set
+            {
+                _previousSelectedMenuOptionIndex = value;
+                OnPropertyChanged("PreviousSelectedMenuOptionIndex");
+                OnPropertyChanged("IsBackButtonEnabled");
+            }
+        }
 
         public Boolean IsBackButtonEnabled
         {
             get
             {
-                return _menuSelectionHistory.Count > 0;
+                return PreviousSelectedMenuOptionIndex.HasValue;
             }
         }
 
@@ -488,8 +498,7 @@ namespace GraphicalEditor
         {
             if (SelectedMenuOptionIndex != ListViewExtendMenu.SelectedIndex)
             {
-                _menuSelectionHistory.Push(SelectedMenuOptionIndex);
-                OnPropertyChanged("IsBackButtonEnabled");
+                PreviousSelectedMenuOptionIndex = SelectedMenuOptionIndex;
                 SelectedMenuOptionIndex = ListViewExtendMenu.SelectedIndex;
             }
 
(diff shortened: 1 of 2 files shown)


VasilijeBursac:

HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -491,6 +511,43 @@ private void ListViewExtendMenu_SelectionChanged(object sender, SelectionChanged
...
+                _selectedMenuOptionsHistory.Add(currentlySelectedMenuOptionIndex);
+            }
+            else
+            {
+                _selectedMenuOptionsHistory.RemoveAt(0);
+                _selectedMenuOptionsHistory.Add(currentlySelectedMenuOptionIndex);
+
+                PreviousSelectedMenuOptionIndex = _selectedMenuOptionsHistory[0];
+                SelectedMenuOptionIndex = _selectedMenuOptionsHistory[1];
+            }
```
milana-todorovic: Could you explain why this method looks like this? Is the goal to limit the number of remembered items, or? I feel like there must be a nicer way to do that.

milana-todorovic (commented):
I think I'm having some issues with the readability here, I would appreciate a quick explanation of how exactly you want the back button to work so I could maybe make some suggestions on how to improve the implementation.

HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml.cs
VasilijeBursac: List _selectedMenuOptionsHistory always contains indexes of the last two selected side panels. Initially, that list is empty, so firstly I just add indexes to the list. Later, on each selection change of side panel index, I just remove the first item from the list because that is now the oldest index, therefor previously selected index will move to the first position in the list and the currently selected will be on the second position in the list. So list always contains just two items: Previously and Currently selected index.

kmilena98 (dismissed):
Looks good and clean

HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml.cs
milana-todorovic: I found what the issue with just using the SelectedMenuItemIndex and adding a field for just the previous index was. I made two suggestions for alternate implementations with that issue fixed, you can check them out over on PR #124 and PR #125. Feel free to do merge or close those PRs as you see fit.

VasilijeBursac: Thank you very much for your time and advice! 😄

VasilijeBursac:
@milana-todorovic  I have tested and changed both One-Step and Multi-Step Back Button implementations to fit our needs (just small corrections I mentioned in comments on your PRs) and committed everything in this PR. We decided to stick with One-Step Back Button but maybe we will change that in the future... 

Thank you very much for your advice, time and effort! We appreciate that! 😄

milana-todorovic (dismissed):
Looks good - although, I just realized it totally could've been implemented while keeping the binding, could've just saved the previous value in the setter for SelectedMenuOptionIndex. I think this is alright as it is though.

VasilijeBursac:
> Looks good - although, I just realized it totally could've been implemented while keeping the binding, could've just saved the previous value in the setter for SelectedMenuOptionIndex. I think this is alright as it is though.

@milana-todorovic  I have changed it now according to your idea and it works. Check out the newest commit when you can. Thanks!

VasilijeBursac:
> Looks good - although, I just realized it totally could've been implemented while keeping the binding, could've just saved the previous value in the setter for SelectedMenuOptionIndex. I think this is alright as it is though.

 I have changed it now according to your idea and it works. Check out the newest commit when you can. Thanks!

milana-todorovic (approved):
Looks good!

kmilena98 (approved):
Very well

PR #122, opened 2020-12-09
feat[g2.2]: Advanced search for objects on the map

VasilijeBursac:
- This branch contains code for displaying selected search result object specially marked and selected on the map.
- Also, it contains logic for switching building floors if the floor that contains desired building objects isn't currently displayed (switching floors automatically).

- Using static attribute _allMapObjects from MainWindow definitely isn't the best option but this will be fixed and a special class will be created at the end of this sprint probably or maybe during next week.


VasilijeBursac:
[(1) g2.2: Advanced search for objects on the map: 5h [3]](https://trello.com/c/rpdorYqc/35-1-g22-advanced-search-for-objects-on-the-map-5h-3)

kmilena98 (approved):
Looks clean and good

milana-todorovic (approved):
Seems alright, as far as I checked it works as intended.

PR #124, opened 2020-12-10
gui: alternate back button implementation for GraphicalEditor

milana-todorovic:
@VasilijeBursac 
I found what the issue with implementing the back button without the list was on PR #121. It doesn't actually have anything to do with references, it was because SelectedMenuItemIndex was bound to the list and was already changed when the selection changed event fired.
Here is an alternate implementation of the back button with that issue fixed for you to consider.

### DIFF
--- a/HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml
+++ b/HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml
@@ -364,7 +364,6 @@
                                     FontSize="18"
                                     Foreground="#FF313131"
                                     ScrollViewer.HorizontalScrollBarVisibility="Disabled"
-                                    SelectedIndex="{Binding SelectedMenuOptionIndex}"
                                     SelectionChanged="ListViewExtendMenu_SelectionChanged">
 
                                     <ListViewItem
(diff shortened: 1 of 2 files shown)


VasilijeBursac:
> @VasilijeBursac
> I found what the issue with implementing the back button without the list was on PR #121. It doesn't actually have anything to do with references, it was because SelectedMenuItemIndex was bound to the list and was already changed when the selection changed event fired.
> Here is an alternate implementation of the back button with that issue fixed for you to consider.

I understand... It was very late when I was implementing this and I was tired so I totally missed that... Thank you for your time and recommendation! I appreciate it!

VasilijeBursac:
This solution definitely works and it is easier to understand for a person who reads it.
Although it has some trade-offs because binding ListView SelectedIndex (I mentioned it in code comments), but that is definitely minor compared to the improvements that this solution brings... I will check the solution with multiple back steps, and then I will choose the final solution. Thank you very much for your time and effort! 😄

VasilijeBursac:

milana-todorovic:
This PR was used as reference for reimplementation on PR #121, so there is no need to merge it.

PR #125, opened 2020-12-10
gui: multi step back button for GraphicalEditor

milana-todorovic:
@VasilijeBursac 
Here is an alternate suggestion for the back button - why not offer the option to go back multiple steps, and not just one? I'm fine with whatever you prefer of course, but it was rather easy to implement after the binding issue was fixed so I thought I'd suggest it.


VasilijeBursac:
> @VasilijeBursac
> Here is an alternate suggestion for the back button - why not offer the option to go back multiple steps, and not just one? I'm fine with whatever you prefer of course, but it was rather easy to implement after the binding issue was fixed so I thought I'd suggest it.

Thank you, I will definitely try both options!

VasilijeBursac:

milana-todorovic:
This PR was used as reference for reimplementation on PR #121, so there is no need to merge it.

PR #140, opened 2020-12-13
feat[g3.1-g3.4]: Grouping appointments by rooms that are received from appointment search

ogikuzma:
- Added 2 methods that will be used by the frontend ( not currently in use)


anciz (dismissed):
Seems to be right.

HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml.cs
```diff
@@ -217,6 +201,32 @@ public MainWindow(string currentUserRole)
...
+                bool found = false;
+                foreach (ExaminationDTO e in appointmentsWithoutDuplicateTimes)
+                {
+                    if (e.DateTime.Equals(appointment.DateTime))
+                        found = true;
+                }
+                if (!found)
+                    appointmentsWithoutDuplicateTimes.Add(appointment);
+            }
+        }
```
milana-todorovic: I assume you meant for this to return the appointmentsWithoutDuplicateTimes though, right? Also, that inner foreach and found bool could be replaced by <code>appointmentsWithoutDuplicateTimes.Any(a => a.DateTime.equals(appointment.DateTime))</code>. No need to reimplement things Linq already does for you.

milana-todorovic (dismissed):
The goal is to group together the appointments that are at the same time, and if there are multiple mark all the rooms that are free at that time, correct? This seems like it will do the job well enough.

HealthcareSystemFrontends/GraphicalEditor/MainWindow.xaml.cs
ogikuzma: Any returns a boolean, whether an object in the list satisfies the condition, right? So, in order to add an appointment to the list or not, depends on what any returns. That will mean I will have one huge if where I will put any statement. Changing inner foreach with the LINQ expression you mentioned would be a nice add-on, but I think it would be better to keep the found bool in order not to have that big if.

milana-todorovic: Yeah, I agree, keeping found with the Linq makes things more readable. I like how it turned out.

Applied in de641e4d5b7e2f7cd1d9f8e983ec99ab5b42d50d.

milana-todorovic (approved):
Seems alright.

anciz (approved):

vlaksi (approved):

PR #149, opened 2020-12-16
refactor[]: change datas(only previous examinations) for search doc

draganacarapic:
* Previously all examinations of one patient could be searched, now I have corrected that only previous examinations are could be searched, because it is a search of examination reports, so it is assumed that these are previous examinations.

### DIFF
--- a/HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
+++ b/HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
@@ -62,9 +62,13 @@ namespace PatientWebApp.Controllers
                 _examinationService.AdvancedSearch(examinationSearchDTO).ForEach(examination => examinationDTOs.Add(ExaminationMapper.ExaminationToExaminationDTO(examination)));
                 return Ok(examinationDTOs);
             }
-            catch (NotFoundException exception)
+            catch (ArgumentNullException)
             {
-                return NotFound(exception.Message);
+                return BadRequest("Patient's jmbg cannot be null.");
+            }
+            catch (DatabaseException e)
+            {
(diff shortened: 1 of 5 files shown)


HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -62,9 +62,13 @@ public ActionResult AdvanceSearchExaminations(ExaminationSearchDTO examinationSe
                 _examinationService.AdvancedSearch(examinationSearchDTO).ForEach(examination => examinationDTOs.Add(ExaminationMapper.ExaminationToExaminationDTO(examination)));
                 return Ok(examinationDTOs);
             }
-            catch (NotFoundException exception)
+            catch (ArgumentNullException)
             {
-                return NotFound(exception.Message);
+                return BadRequest("Patient's jmbg cannot be null.");
```
milana-todorovic, line 67: Why does argument null exception happen here and mean this? Can you explain where in the stack this happens?

draganacarapic: When I changed in the service in the method for search calls a method that returns only previous examinations, NotFoundException no longer made sense, because the newly called method does not throw NotFoundException. 
Then I forwarded null to the patientJmbg and saw it being treated as ArgumentNullException, so I just changed that for that case I catch this exception.

milana-todorovic: Alright, although it would make more sense to just validate the DTO and not call the method if it isn't correct.

milana-todorovic (approved):

sladjana-savkovic (approved):

PR #150, opened 2020-12-18
refactor[w3.1,w3.2]:setting limits for search result

sladjana-savkovic:
I want to set a limit to the results for basic appointment search (the system should return a maximum of ten free appointments). The reason is faster delivery of free appointments.

### DIFF
--- a/HealthcareSystem/PatientWebApp/Controllers/AppointmentController.cs
+++ b/HealthcareSystem/PatientWebApp/Controllers/AppointmentController.cs
@@ -8,6 +8,7 @@ using Backend.Service;
 using Backend.Service.ExaminationAndPatientCard;
 using Microsoft.AspNetCore.Mvc;
 using Model.PerformingExamination;
+using Model.Users;
 using PatientWebApp.DTOs;
 using PatientWebApp.Mappers;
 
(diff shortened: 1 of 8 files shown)


HealthcareSystem/PatientWebApp/Controllers/AppointmentController.cs
```diff
@@ -72,6 +75,8 @@ public IActionResult PrioritySearchAppointments(AppointmentSearchWithPrioritiesD
                 parameters.InitialParameters.IsAppointmentValid();
                 freeAppointments = _freeAppointmentSearchService.SearchWithPriorities(parameters).ToList();
                 freeAppointments.ForEach(appointment => freeAppointmentsDTOs.Add(ExaminationMapper.ExaminationToExaminationDTO(appointment)));
+                freeAppointmentsDTOs.ForEach(appointmentDTO => appointmentDTO.DoctorName = _doctorService.ViewProfile(appointmentDTO.DoctorJmbg).Name);
+                freeAppointmentsDTOs.ForEach(appointmentDTO => appointmentDTO.DoctorSurname = _doctorService.ViewProfile(appointmentDTO.DoctorJmbg).Surname);
                 return Ok(freeAppointmentsDTOs);
```
milana-todorovic: I really wouldn't call the service method twice, you're making to database queries for the same data here. Just save the Doctor in a temporary variable. Also, why is the method in the DoctorService called ViewProfile? Could it be changed to Get?

milana-todorovic (approved):
Seems alright.

HealthcareSystem/PatientWebApp/Controllers/AppointmentController.cs
milana-todorovic: Fixed in a0b2c915b5a87cfd127459a5fc3f21d216caee18.

jelenazeljko55 (approved):

PR #157, opened 2020-12-23
feat[g3.2]: Doctor priority search enhancement

ogikuzma:
Doctor priority search enhancement:
- Optimization for appointment search ( +- 1 day step)
- Check if the date has already passed while making an interval for doctor priority search
- One test in WebAppTests is commented because new logic produces 34 instead of 10 free appointments as it was placed in the test


ogikuzma:
I've just done a little bit more testing and from what I can see search could take a bit longer than expected. So, if we decide to optimize the method even more this method GetPotentiallyAvailableAppointments should be tackled since it takes most of the time from the whole search.

sladjana-savkovic (approved):

anciz (approved):
Seems to be right.

Pufke (approved):

PR #158, opened 2020-12-23
feature-w4.1-login-user

sladjana-savkovic:
The goals:
-  Create service and repository layers for admin and add admin in data seeder
-  Make a controller for login user
-  Get user's credentials from request body, generate json web token and save it to local storage
-  Implement user authentication logic
-  Implement user authorization logic


sladjana-savkovic:
[(8) W4.1: As an user I want to be able to login into application, so that I could use hospital services.](https://trello.com/c/PsEufA1T/36-8-w41-as-an-user-i-want-to-be-able-to-login-into-application-so-that-i-could-use-hospital-services)

vlaksi (dismissed):

vlaksi (dismissed):

anciz (dismissed):

HealthcareSystem/PatientWebApp/Controllers/UserController.cs
```diff
@@ -0,0 +1,137 @@
...
+               ValidateAudience = false
+           };
+           var claims = handler.ValidateToken(token, validations, out var tokenSecure);
+
+           string username = claims.Claims.First(c => c.Type == ClaimTypes.NameIdentifier).Value;
+           string role = claims.Claims.First(c => c.Type == ClaimTypes.Role).Value;
+           string jmbg = claims.Claims.First(c => c.Type == "Jmbg").Value;
+
+           return Ok(new LoggedUserDTO(username, jmbg, role));
+       }
```
milana-todorovic: What is the purpose of this method? Are you sending your JWT to the backend to fetch data every time? Why not just put the data you need in the JWT and read it from Javascript? Data you need in the frontend, that is. For AJAX just leave out any data that depends on the logged in user, send the JWT in the Authorization header, and use the identity you get from the JWT to fill in the missing data on the backed.

HealthcareSystem/PatientWebApp/wwwroot/js/add_feedback.js
```diff
@@ -20,9 +46,9 @@
 			return;
 		}
 
-		//I take the certain patient from the database, otherwise the currently logged in patient will be taken
+		//I take the certain patient 1309998775018 from the database, otherwise the currently logged in patient will be taken
```
milana-todorovic: Why do this? Shouldn't you just use the JMBG of whoever is logged in? Or rather, you shouldn't send a JMBG at all, you should send the JWT and the backend should get the JMBG from that as I mentioned above.

HealthcareSystem/PatientWebApp/wwwroot/js/unpublished_feedback.js
```diff
@@ -1,4 +1,30 @@
...
+			dataType: 'json',
+			processData: false,
+			contentType: 'application/json',
+			data: JSON.stringify(token),
+			success: function (loggedUser) {
+				if (loggedUser.role != "Admin") {
+					alert('Access denied!');
+					return;
+				}
+				jmbg = loggedUser.jmbg;
```
milana-todorovic: Refer above to what I said about JMBG, fetching it here is odd and saving it in a variable that you later use for AJAX is even odder. But also, one this that also jumps out at me since I see you doing this everywhere - I feel like you should just make one JS file with all the JWT utility methods you'll need everywhere (things like, check if there's a JWT saved and redirect if no one is logged in or the role is wrong, get the Bearer string for when you need it for AJAX, and so on), and just include that JS file and use it wherever you need it.

HealthcareSystem/PatientWebApp/wwwroot/js/filling_out_the_survey.js
```diff
@@ -1,7 +1,32 @@
...
+				}
+				jmbg = loggedUser.jmbg;
+			},
+			error: function () {
+				alert('Error getting logged user!');
+			}
+		});
+	}
+	else {
+		alert('Unlogged user!');
```
milana-todorovic: Also, please remove calls to alert before the branch is merged. In these cases I feel like a redirect to the login page would be more appropriate anyway (since you're not supposed to be able to access these pages without being logged in).

HealthcareSystem/PatientWebApp/wwwroot/js/add_feedback.js
jelenabudisa: I corrected your suggestions except this one. If other parts are alright, my next commit will be changing AJAX calls so that they have a JWT send through a header to backend (not a JMBG through an URL).

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -33,6 +36,8 @@ public ExaminationController(IExaminationService examinationService)
         /// </summary>
         /// <param name="patientJmbg">patients jmbg</param>
         /// <returns>if alright returns code 200(Ok), if not 404(not found)</returns>
+        /// 
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpGet("by-patient/{patientJmbg}")]
         public ActionResult GetExaminationsByPatient(string patientJmbg)
         {
```
milana-todorovic: I think this endpoint should be transformed to get the jmbg from the Principal like GetPreviousExaminationsByPatient does.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -53,6 +58,8 @@ public ActionResult GetExaminationsByPatient(string patientJmbg)
         /// </summary>
         /// <param name="examinationSearchDTO">an object need be find in the database</param>
         /// <returns>if alright returns code 200(Ok), if not 400(not found)</returns>
+        /// 
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpPost("advance-search")]
         public ActionResult AdvanceSearchExaminations(ExaminationSearchDTO examinationSearchDTO)
         {
```
milana-todorovic, line 66: This too, I think the DTO contains the jmbg, but you should get it from the Principal instead.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -78,6 +85,8 @@ public ActionResult AdvanceSearchExaminations(ExaminationSearchDTO examinationSe
         /// </summary>
         /// <param name="id">id of the object to be changed</param>
         /// <returns>if alright returns code 200(Ok), if not 400(bed request), if connection lost returns 500</returns>
+        /// 
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpPut("cancel/{id}")]
         public ActionResult CancelExamination(int id)
         {
```
milana-todorovic: You also need a jmbg check here, patients can only cancel their own examinations.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -178,6 +196,8 @@ public ActionResult GetFollowingExaminationsByPatient(string patientJmbg)
         /// </summary>
         /// <param name="id">examination id</param>
         /// <returns>if alright returns code 200(Ok), if returned value is null returns 404, if connection lost returns 500</returns>
+        /// 
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpGet("{id}")]
         public IActionResult GetExaminationById(int id)
```
milana-todorovic, line 210: Jmbg check here as well after you fetch the examination, it needs to be for this patient.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -202,6 +222,8 @@ public IActionResult GetExaminationById(int id)
         /// </summary>
         /// <param name="examinationDTO">an object to be added to the database</param>
         /// <returns>if alright returns code 201(Created), if validation isn't successful return 400, if connection lost returns 500</returns>
+        /// 
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpPost]
         public IActionResult AddExamination(ExaminationDTO examinationDTO)
         {
```
milana-todorovic, line 243: Here too, the patient can't schedule examinations for other patients.

HealthcareSystem/PatientWebApp/Controllers/FeedbackController.cs
```diff
@@ -48,6 +52,8 @@ public IActionResult GetFeedbackById(int id)
         /// </summary>
         /// <param name="feedbackDTO">an object to be added to the database</param>
         /// <returns>if alright returns code 200(Ok), if not 400(bed request)</returns>
+        /// 
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpPost]
         public ActionResult AddFeedback(FeedbackDTO feedbackDTO)
         {
```
milana-todorovic, line 40: You should probably also check here that the jmbg in the DTO matches the principal.

HealthcareSystem/PatientWebApp/Controllers/FeedbackController.cs
```diff
@@ -30,6 +32,8 @@ public FeedbackController(IFeedbackService feedbackService)
         /// </summary>
         /// <param name="id">id of the wanted object</param>
         /// <returns>if alright returns code 200(Ok), if not 404(not found)</returns>
+        /// 
+        [AllowAnonymous]
         [HttpGet("{id}")]
         public IActionResult GetFeedbackById(int id)
         {
```
milana-todorovic: Does this only return the published or all? Only admin should see the unpublished ones. (Although, is this endpoint ever used anyway? I can't remember any use case that requires you to fetch only one feedback.)

HealthcareSystem/PatientWebApp/Controllers/PatientController.cs
```diff
@@ -45,11 +48,14 @@ public PatientController(IPatientService patientService, IPatientCardService pat
...
-        public IActionResult GetPatientByJmbg(string jmbg)
+        /// 
+        [Authorize(Roles = UserRoles.Patient + "," + UserRoles.Admin)]
+        [HttpGet]
+        public IActionResult GetPatientByJmbg()
         {
             try
             {
+                var jmbg = HttpContext.User.FindFirst("Jmbg").Value;
                 Patient patient = _patientService.ViewProfile(jmbg);
```
milana-todorovic: Does admin ever need this? I think the admin only views malicious patients.

HealthcareSystem/PatientWebApp/Controllers/SurveyController.cs
```diff
@@ -33,7 +36,7 @@ public SurveyController(ISurveyService surveyService, IExaminationService examin
             _examinationValidator = new ExaminationValidator(_examinationService);
         }
 
-
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpPost]
         public ActionResult AddSurvey(SurveyDTO surveyDTO)
         {
```
milana-todorovic, line 43: Patients can only fill out the survey for their own examinations, that needs to be checked as well.

HealthcareSystem/PatientWebApp/Controllers/TherapyController.cs
```diff
@@ -30,6 +33,8 @@ public TherapyController(ITherapyService therapyService)
         /// </summary>
         /// <param name="patientJmbg">patients jmbg</param>
         /// <returns>if alright returns code 200(Ok), if not 404(not found)</returns>
+        /// 
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpGet("{patientJmbg}")]
         public ActionResult GetTherapiesByPatient(string patientJmbg)
         {
```
milana-todorovic: Get the jmbg from the Principal instead.

HealthcareSystem/PatientWebApp/Controllers/TherapyController.cs
```diff
@@ -50,6 +55,8 @@ public ActionResult GetTherapiesByPatient(string patientJmbg)
         /// </summary>
         /// <param name="therapySearchDTO">an object need be find in the database</param>
         /// <returns>if alright returns code 200(Ok), if not 400(not found)</returns>
+        /// 
+        [Authorize(Roles = UserRoles.Patient)]
         [HttpPost("advance-search")]
         public ActionResult AdvanceSearchTherapies(TherapySearchDTO therapySearchDTO)
         {
```
milana-todorovic, line 63: Here as well, remove the jmbg from the DTO and get it from the principal.

HealthcareSystem/PatientWebApp/DTOs/LoggedUser.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace PatientWebApp.DTOs
+{
+    public class LoggedUserDTO
+    {
```
milana-todorovic: I think this class is no longer used anywhere.

HealthcareSystem/PatientWebApp/wwwroot/js/activate_account.js
```diff
@@ -10,7 +10,7 @@
             success: function () {
 
                 setTimeout(function () {
-                    window.location.href = 'patients_home_page.html';
+                    window.location.href = 'login.html';
                 }, 500);
```
milana-todorovic, line 14: Maybe a good improvement for the future could be to make the activation endpoint do login automatically as well.

milana-todorovic (commented):
Looks good, you just need to tighten the authorization a tiny bit for some endpoints. It should be quick work though, you already have similar logic figured out and in place on other endpoints.

HealthcareSystem/PatientWebApp/wwwroot/js/activate_account.js
draganacarapic: I think that our product owner requested redirection to the login page after the successful activation of the account.

jelenabudisa (commented):
Some tests broke down due to massive changes. We will keep them in comments for now. In the future, we shall have a huge revise of the tests.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -87,11 +90,16 @@ public ActionResult AdvanceSearchExaminations(ExaminationSearchDTO examinationSe
...
-        [HttpPut("cancel/{id}")]
+        [HttpPost("cancel/{id}")]
         public ActionResult CancelExamination(int id)
         {
             try
             {
+                var patientJmbg = HttpContext.User.FindFirst("Jmbg").Value;
+                Examination examination= _examinationService.GetExaminationById(id);
+                if (!patientJmbg.Equals(examination.PatientCard.PatientJmbg)) {
+                    return StatusCode(400, "Patient tried to cancel someone else's examination");
```
milana-todorovic: This should probably be 403.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
```diff
@@ -203,6 +211,12 @@ public IActionResult GetExaminationById(int id)
         {
             try
             {
+                var patientJmbg = HttpContext.User.FindFirst("Jmbg").Value;
+                Examination examination = _examinationService.GetExaminationById(id);
+                if (!patientJmbg.Equals(examination.PatientCard.PatientJmbg))
+                {
+                    return StatusCode(400, "Patient tried to get someone else's examination");
```
milana-todorovic: 403 here as well.

milana-todorovic (dismissed):
Seems alright, I would just recommend 403 instead of 400 for the situations where patients try to access another patient's data.
And I'm fine with commenting out the tests for now, we'll have to look at all our current tests anyway and see what we're keeping, moving, or getting rid of because of the system architecture changes that are planned for this sprint.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
milana-todorovic: Fixed.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
milana-todorovic: Fixed.

HealthcareSystem/PatientWebApp/Controllers/ExaminationController.cs
milana-todorovic: Fixed.

HealthcareSystem/PatientWebApp/Controllers/FeedbackController.cs
milana-todorovic: Fixed.

HealthcareSystem/PatientWebApp/Controllers/SurveyController.cs
```diff
@@ -44,6 +48,8 @@ public ActionResult AddSurvey(SurveyDTO surveyDTO)
                 _examinationValidator.CheckIfSurveyAboutExaminationIsCompleted(surveyDTO.ExaminationId);
                 _surveyService.AddSurvey(SurveyMapper.SurveyDTOToSurvey(surveyDTO));
                 _examinationService.CompleteSurveyAboutExamination(surveyDTO.ExaminationId);
+                if(!_examinationService.GetExaminationById(surveyDTO.ExaminationId).PatientCard.PatientJmbg.Equals(HttpContext.User.FindFirst("Jmbg").Value))    
+                    return BadRequest("Patient can only fill out the survey for their own examinations.");
```
milana-todorovic: This should happen before the service is called (but all these methods will be refactored soon so it can be moved around then).

HealthcareSystem/PatientWebApp/Controllers/TherapyController.cs
milana-todorovic: Fixed.

milana-todorovic (approved):

draganacarapic (approved):

PR #161, opened 2020-12-24
gui[w4.2]: redesign home pages

draganacarapic:
* I redesigned the home pages for the patient(preview pharmacy actions) and for admin(publishing feedback).


draganacarapic:
[(3) W4.2: As a developer I want to design home pages for patients and system administrators, so that I could redirect them after successful login.](https://trello.com/c/3vuE3lyd/37-3-w42-as-a-developer-i-want-to-design-home-pages-for-patients-and-system-administrators-so-that-i-could-redirect-them-after-succ)

vlaksi (dismissed):

anciz (dismissed):

HealthcareSystem/PatientWebApp/DTOs/PharmacyDTO.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class PharmacyDTO
+    {
+        public int Id { get; set; }
+        public string Name { get; set; }
+        public string ApiKey { get; set; }
+        public string Url { get; set; }
+        public string ActionsBenefitsExchangeName { get; set; }
+        public bool ActionsBenefitsSubscribed { get; set; } = false;
+        public string GrpcHost { get; set; }
+        public int GrpcPort { get; set; }
```
milana-todorovic: Why would you expose any of these to these in the PatientWebApp API? Why not just put the Pharmacy name in the ActionBenefitDTO instead of the Id and put it together when getting them? You just need the name anyway, and some things you're sending here are potentially sensitive data that shouldn't be exposed to just anyone like this.

HealthcareSystem/PatientWebApp/DTOs/ActionBenefitDTO.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace PatientWebApp.DTOs
+{
+    public class ActionBenefitDTO
+    {
+        public int Id { get; set; }
+        public int PharmacyId { get; set; }
+        public string Subject { get; set; }
+        public string Message { get; set; }
+        public bool IsPublic { get; set; }
```
milana-todorovic: Why would you need IsPublic in the DTO? It's not like the patient should be able to see any that aren't public.

milana-todorovic (commented):
Fix the DTOs. They should contain the data you actually need, not any and all data you have. Also, try not to make multiple AJAX calls for things like these, putting the data together on the backend during your first call is usually easier and saves both time and bandwidth.

milana-todorovic (commented):
Also, a practical suggestion for showing these on the homepage. Having three fixed slides is odd and impractical, obviously. I think you should have one fixed slide that says something like 'Actions and benefits' or 'Promotions' or whatever, mostly just to have something to show when you have no promotions, and then insert extra slides for every promotion. Having like 3 images is fine, just keep a count as you add the slides and cycle through the images with mod.

milana-todorovic (approved):
Looks good!

HealthcareSystem/PatientWebApp/DTOs/PharmacyDTO.cs
milana-todorovic: Fixed in a5ad64709381a12cb71e9eaa5b25cbba14c76a01.

HealthcareSystem/PatientWebApp/DTOs/ActionBenefitDTO.cs
milana-todorovic: Fixed in a5ad64709381a12cb71e9eaa5b25cbba14c76a01.

anciz (approved):

PR #163, opened 2020-12-28
refactor-w4.3-patient-service

sladjana-savkovic:
The goals:
-  Implement methos in PatientAccount and PatientService
-  Create PatientController
-  Connect PatientControllers from PatientWebApp and UserService through HTTP

### DIFF
--- a/HealthcareSystem/UserService/Model/PatientAccount.cs
+++ b/HealthcareSystem/UserService/Model/PatientAccount.cs
@@ -23,6 +23,7 @@ namespace UserService.Model
         {
             if (!IsActivated) throw new ValidationException("Blocking unactivated patient account isn't possible.");
             if (IsBlocked) throw new ValidationException("Patient account is already blocked.");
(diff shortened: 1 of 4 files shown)


HealthcareSystem/UserService/Model/PatientAccount.cs
```diff
@@ -1,35 +1,45 @@
...
+            IsActivated = true;
         }
 
         public void Block()
         {
-            throw new NotImplementedException();
+            if (!IsActivated) throw new ValidationException("Blocking unactivated patient account isn't possible.");
+            if (IsBlocked) throw new ValidationException("Patient account is already blocked.");
+
+            IsBlocked = true;
```
milana-todorovic: Should you check if the patient is malicious here as well? Can patients that aren't malicious be blocked?

sladjana-savkovic: You're right, I forgot that condition.

sladjana-savkovic:

milana-todorovic (approved):

PR #164, opened 2020-12-29
refactor[w4.3]: doctor service

jelenazeljko55:


jelenazeljko55:

HealthcareSystem/UserService/Model/Email.cs
```diff
@@ -25,7 +26,16 @@ public override int GetHashCode()
...
         private void Validate()
         {
-            throw new NotImplementedException();
+            if (IsRegular(Value))
+            {
+                throw new Exception();
+            }
+        }
+
+        private bool IsRegular(string Value)
```
milana-todorovic: I don't really like this name, I feel like it doesn't tell me anything. Maybe something more specific would be better.

HealthcareSystem/UserService/Model/DoctorAccount.cs
```diff
@@ -13,7 +14,7 @@ protected override void Validate()
             base.Validate();
             if (DateOfEmployment.CompareTo(DateTime.Now) > 0)
             {
-                throw new Exception("Date of employment not valid.");
+                throw new ValidationException("Date of employment not valid.");
             }
```
milana-todorovic: What did we decide in the end, are we checking dates or avoiding it because of the DateTime.Now dependency?

HealthcareSystem/UserService/Model/Email.cs
```diff
@@ -26,10 +27,11 @@ public override int GetHashCode()
 
         private void Validate()
         {
-            if (!IsRegular(Value)) throw new Exception("Invalid email");
+            if (string.IsNullOrEmpty(Value)) throw new ValidationException("Email does not exist!");
+            if (!IsGoodFormat(Value)) throw new ValidationException("Invalid email");
         }
 
-        private bool IsRegular(string Value)
+        private bool IsGoodFormat(string Value)
```
milana-todorovic: @draganacarapic named a similar method IsValidFormat or something like that on another branch, one of you should change so it matches.

HealthcareSystem/UserService/Model/Jmbg.cs
```diff
@@ -26,10 +27,11 @@ public override int GetHashCode()
 
         private void Validate()
         {
-            if (!IsRegular(Value))  throw new Exception("Invalid jmbg");       
+            if (string.IsNullOrEmpty(Value)) throw new ValidationException("Jmbg does not exist!");
```
milana-todorovic: I wouldn't say 'doesn't exist' here, maybe something like 'Jmbg can't be empty' or 'Jmbg is required' is more accurate?

HealthcareSystem/UserService/Model/Specialty.cs
```diff
@@ -19,7 +20,10 @@ public Specialty(int id, string name)
 
         private void Validate()
         {
-            throw new NotImplementedException();
+            if (string.IsNullOrEmpty(Name))
+            {
+                throw new ValidationException("Specialty does not exist!");
```
milana-todorovic: Here as well, I don't like 'does not exist' as the message for empty names.

milana-todorovic (approved):

PR #166, opened 2020-12-29
refactor[w4.3]: user service

draganacarapic:


draganacarapic:

HealthcareSystem/UserService/Model/Password.cs
```diff
@@ -25,7 +27,13 @@ public override int GetHashCode()
 
         private void Validate()
         {
-            throw new NotImplementedException();
+            if (string.IsNullOrEmpty(Value)) throw new ValidationException("Password cannot be null or empty.!");
+            if (!IsValidFormat(Value)) throw new ValidationException("Invalid password.");
+        }
+        private bool IsValidFormat(string Value)
+        {
+            Regex regex = new Regex(@"^[a-zA-Z0-9\.\-_]{8,20}$");
```
milana-todorovic: Why wouldn't we allow other special characters in the password though? I'd probably just like, do a trim and check if the length is alright. 
Although, I just realized that we did something really stupid with this bit of the model, specifically that if this check goes here, we can't encrypt the password. So I'd discuss this at the next meeting, I would maybe toss this class entirely and just check that the password isn't null or empty in User. And then I'd move the whole logic for the password length and such into the API Gateway, since it's more of a security concern and that's where we're doing security right now? And then I'd encrypt the password in the API Gateway before sending it anywhere further.

HealthcareSystem/UserService/Model/MaliciousAction.cs
```diff
@@ -19,7 +20,7 @@ public MaliciousAction(MaliciousActionType type, DateTime timeStamp)
 
         private void Validate()
         {
-            throw new NotImplementedException();
+            if (!Type.Equals(MaliciousActionType.AppointmentCancellation)) throw new ValidationException("Type of malicious action must be appointment cancellation.");
```
milana-todorovic: Eh, I'd delete this, it's not like it can be anything else. And if we add something else we'd have to change this, so it's just a pain.

draganacarapic: Should I delete the method: Validate()?

milana-todorovic (approved):

PR #177, opened 2021-01-06
refactor[w4.3]:connecting controllers from patient service

draganacarapic:
* I will implement the method for preview of patient card when User service will connect to the develop.


draganacarapic:

PR #189, opened 2021-01-09
chore[d70]: delivery pipeline maintenance

milana-todorovic:
- Services added to test env on Heroku.
- Existing delivery pipelines reconfigured and pipelines added for new components. The new delivery pipelines will only redeploy components that have been changed. Database reset and E2E tests will run on every commit on develop.
- Simple ping endpoints added to all servers for smoke testing.
- Data seeding changed to work through SQL scripts generated from a DbContext. This decision was made because in the previous approach one of the components had to be responsible for data seeding, which meant that it had to be redeployed whenever we wanted to reset the database, and also that data seeding was attempted every time Heroku restarted the container.
- Data seeding was also changed so that EF migrations and script generation can be invoked through Backend instead of through the ASP.NET apps, with the goal to make script generation faster. This was achieved by adding the classes PostgreSeededDbContext and MySqlSeededDbContext which provide empty constructors so that EF can instantiate them. This implementation is very far from ideal, however, as far as I'm aware we have no other option since EF Core 3.1 has a bug that makes context factories ignore arguments that are passed in (which means that we wouldn't be able to generate both scripts for Postgre and for MySql).
- The E2E pipeline is set up but isn't quite finished yet (since we don't have E2E tests right now). It will be properly finished on another branch when E2E tests are added.

### DIFF
--- a/HealthcareSystem/Backend/Model/DataSeededDbContext.cs
+++ b/HealthcareSystem/Backend/Model/DataSeededDbContext.cs
@@ -14,14 +14,11 @@ using System.Text;
 
 namespace Backend.Model
 {
-    class TestDbContext : MyDbContext
+    class DataSeededDbContext : MyDbContext
     {
-        private static readonly DbContextOptions<MyDbContext> options =
-            new DbContextOptionsBuilder<MyDbContext>().UseNpgsql(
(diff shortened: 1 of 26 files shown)


vlaksi (approved):

anciz (approved):
Alright!

PR #190, opened 2021-01-09
test[w1.1, w1.3 , w3.5, w3.7]: implement end to end tests

draganacarapic:
* Start work on End to end test implementation


draganacarapic:

milana-todorovic (dismissed):
Could you change the test so it uses FirefoxDriver instead of Chrome? That makes it easier to run the test in the pipeline I believe (or maybe Chrome tests can't run at all on Ubuntu agents?). @Pufke what was the exact issue with Chrome in E2E tests?

Also, since you made a new project for E2E tests, you can delete the old E2ETests projects that we used for testing purposes last sprint to avoid confustion.

milana-todorovic:
The E2E tests haven't been added to the pipeline yet because of some technical issues we're having with Azure DevOps. A new branch will be made for fixing the pipeline.

milana-todorovic (approved):
The tests look good to me.

sladjana-savkovic (approved):

PR #192, opened 2021-01-10
refactor[g4.2-g4.3]: Editing of EmergencyDTO to fit frontend needs and small changes of search logic

ogikuzma:
- Refactor of EmergencyExaminationDTO to fit frontend needs since frontend needs examinations from 2 lists to be in one and sorted
- Deletion of appointment uniqueness for the same DateTime since I realized it isn't actually right


milana-todorovic (approved):
Seems okay.

kmilena98 (approved):
Works well.

PR #194, opened 2021-01-10
feature[d64]: event sourcing service

vlaksi:
- created an event store that stores various types of events
- added the service to CI
- added the service to CD


HealthcareSystem/EventSourcingService/Service/EventStoreExampleService.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public void Add(ExampleEvent exampleEvent)
+        {
+            _exampleEventRepository.Add(exampleEvent);
+        }
+
+        public IEnumerable<ExampleEvent> GetAll(Expression<Func<ExampleEvent, bool>> condition)
+        {
+            throw new NotImplementedException();
+        }
```
milana-todorovic: I don't really thing the service example should have this method, since I figure actual event services won't actually expose this method? They'll just use the matching repository method while making reports.

HealthcareSystem/EventSourcingService/Startup.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddControllers().AddNewtonsoftJson(s => {
+                s.SerializerSettings.ContractResolver = new CamelCasePropertyNamesContractResolver();
+            });
+
+            services.AddDbContext<EventSourcingDbContext>(options =>
+            {
+                options.UseMySql(Configuration.GetConnectionString("DefaultConnection"));
+            });
```
milana-todorovic: Configure this to use Postgres on Test env so we can deploy it.

milana-todorovic (changes requested):
Seems alright, but it needs to be configured so that it can be deployed to Heroku.

pipelines/cd/event-sourcing-service-delivery.yml
```diff
@@ -0,0 +1,82 @@
...
+        tags: |
+          $(Build.BuildId)
+          latest
+
+- stage: CD
+  variables:
+    HEROKU_API_KEY: 8418d6b4-7423-404b-81cf-23abcd968591
+  jobs:
+  - deployment: EventSourcingService  
+    environment: 'Test'
```
milana-todorovic, line 51: We should decide what to do with the test env event database, I would maybe reset it in this pipeline as well just so we don't have issues with missing tables and such if the schema is changed?

milana-todorovic (approved):
Looks good!

Pufke (approved):

Pufke:
3.6

PR #198, opened 2021-01-10
test[w4.3]: add integration tests for user service

draganacarapic:


anciz (dismissed):

anciz (dismissed):

vlaksi (dismissed):

HealthcareSystem/UserServiceTests/IntegrationTests/DbContextInMemory.cs
```diff
@@ -0,0 +1,68 @@
...
+        private DbContextOptionsBuilder<MyDbContext> _builder = new DbContextOptionsBuilder<MyDbContext>();
+        private DbContextOptions<MyDbContext> _options;
+        public MyDbContext _context { get; private set; }
+        public DbContextInMemory()
+        {
+            _builder.UseInMemoryDatabase(Guid.NewGuid().ToString());
+            _options = _builder.Options;
+
+        }
+        public void SeedDatas()
```
milana-todorovic: Data doesn't have a plural.

HealthcareSystem/UserServiceTests/IntegrationTests/DbContextInMemory.cs
```diff
@@ -0,0 +1,68 @@
...
+                DateOfBirth = new DateTime(1965, 11, 08),
+                Phone = "021954201",
+                Email = "milic_milan@gmail.com",
+                HomeAddress = "Aleja Svetog Save 100",
+                Username = "milic_milan@gmail.com",
+                CityZipCode = 3,
+                Gender = GenderType.M,
+                Password = "milanmilic965"
+            });
+            _context.SaveChanges();
```
milana-todorovic: This works okay but maybe using the existing DataSeeder or the SeededDataContext class instead would'be been easier? Since we already have pretty stable data seeding there which I think can work nicely for integration tests as well.

HealthcareSystem/UserServiceTests/IntegrationTests/UserServiceTests.cs
```diff
@@ -0,0 +1,58 @@
...
+    public class UserServiceTests
+    {
+        private UserService.Service.Implementation.UserService SetupRepositoryAndService()
+        {
+            DbContextInMemory testData = new DbContextInMemory();
+            testData.SeedDatas();
+            MyDbContext context = testData._context;
+            var userRepo = new UserRepository(context);
+            return new UserService.Service.Implementation.UserService(userRepo);
+        }
```
milana-todorovic: I feel like returning IUserService here would've been better, just to make the tests a bit less rigid?

HealthcareSystem/UserServiceTests/IntegrationTests/UserServiceTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            UserService.Service.Implementation.UserService userService = SetupRepositoryAndService();
+            try
+            {
+                UserAccount userAccount = userService.GetByEmailAndPassword("bla bla", "bla");
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex is NotFoundException);
+            }
```
milana-todorovic: Do Assert.Throws<NotFoundException>(userService.GetByEmailAndPassword("bla bla", "bla") instead.

HealthcareSystem/UserServiceTests/IntegrationTests/UserServiceTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            Assert.False(userAccount is null);
+        }
+
+        [Fact]
+        public void SuccessLoginAdmin()
+        {
+            UserService.Service.Implementation.UserService userService = SetupRepositoryAndService();
+            UserAccount userAccount = userService.GetByEmailAndPassword("milic_milan@gmail.com", "milanmilic965");
+
+            Assert.False(userAccount is null);
```
milana-todorovic, line 49: I think you should maybe Assert that the correct data is returned here, not just that it's not null.

milana-todorovic (dismissed):
Seems generally alright, I suggested some improvements. Also, maybe putting these tests in a theory would be a good idea? Maybe write one theory that tests whether exceptions are thrown at the right time, one theory that tests if the role is correct if a user is found, one that asserts that the jmbg is correct if the user is found, and so on? So if something fails it'll be easier to tell what exactly is wrong.

milana-todorovic (approved):

HealthcareSystem/UserServiceTests/IntegrationTests/DbContextInMemory.cs
milana-todorovic: Changed in 5e427a9e0a97c3e3c7cb8f0ad3372d302be9569a.

anciz (approved):

PR #202, opened 2021-01-11
feat[g3.1-g4.5-g4.6]:Added checks for scheduled equipmentTransfers

djura023:
-Added checks for  scheduled equipment transfers appointments while searching for avalible examiantion appointments 
-Added checks in scheduling equipment transfer appointment


milana-todorovic (dismissed):
Seems okay. More tests seem to be failing than usual though, you should look at that to make sure if the changes on this branch are causing that, or just the ongoing issues with tests being sensitive to time.

anciz (dismissed):

milana-todorovic (commented):
6 tests still failed. As far as I'm aware only 1 test had the existing timing issue, so does this branch still have issues?

djura023:
> issues

no issues,i forgot to change all tests

milana-todorovic (approved):
Seems alright. The check which didn't finish is because of some issue we're having with Azure DevOps today, we're getting really odd errors with fetching agents. I've tried cancelling and rerunning the check but it didn't really work as expected, the run won't cancel properly. I can merge the branch despite that though, if you've finished everything you planned to do here?

anciz (approved):

PR #205, opened 2021-01-13
test[ia2]: additional tests

borislav-puzic:


milana-todorovic (commented):
A test is failing, I'm assuming it's one of the tests added on this branch?

milana-todorovic (commented):
As far as I can see your ISA project needs to be running in order for these tests to work, and they also potentially depend on the contents of the ISA database. Were you specifically instructed to test like this by your Product Owner? Because it is potentially a huge issue, the ISA project is not a managed dependency, it's a whole external system. We (formally speaking) don't control how, when, and where it gets deployed, or what data it has in it's database at any given time.
If these tests need to stay like this, then we should discuss how to handle them in the pipeline, and whether or not we need to deploy the ISA project in the pipeline (and your new services later as well).

Stefan-Kitanovic (approved):

milana-todorovic (approved):
Approved so that it can be modified to fit with the microservices, and since the product owner agreed that testing this way is acceptable. We'll discuss in the future whether or not the pipeline needs to be modified to fit.

PR #208, opened 2021-01-15
refactor[w5.4]: add steps in scheduling

draganacarapic:


draganacarapic:
[(13) W5.4 : As an admin, I want to observe events regarding process for making new appointment, so that I could find out if our users use the application successfully.](https://trello.com/c/8XTeolJT/42-13-w54-as-an-admin-i-want-to-observe-events-regarding-process-for-making-new-appointment-so-that-i-could-find-out-if-our-users-u)

HealthcareSystem/PatientWebApp/wwwroot/html/patient_examinations.html
```diff
@@ -112,39 +77,66 @@ <h4 class="card-title">Schedule an examination</h4>
...
+                        </select><br />
+                        <button class="btn btn-warning" style="margin-left:110px;" id="step_previous_btn" onclick="step_previous(2)">Previous</button>
+                        <button class="btn btn-success" style="margin-left:20px;" id="step_btn" onclick="step(3)">Next</button>
+                    </div>
+                    <div id="div_appointments" style="display:none;">
+                        <label for="dateOfExam"><b>Available appointments:</b></label>
+                        <select class="form-control" id="free_appointments">
+                        </select><br />
+                        <button class="btn btn-warning" style="margin-left:90px;" id="step_previous_btn" onclick="step_previous(3)">Previous</button>
+                        <button class="btn btn-success" style="margin-left:20px;" id="modalButton" onclick="scheduleExamination()">Schedule</button>
```
milana-todorovic: If you put ids on html elements, make them unique. Not sure why these even need ids though?

HealthcareSystem/PatientWebApp/wwwroot/js/patient_examinations.js
```diff
@@ -386,4 +328,102 @@ function cancelExamination(id) {
...
+            let alert = $('<div class="alert alert-danger alert-dismissible fade show mb-0 mt-2" role="alert">You must select a date in the first step.'
+                + '<button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>' + '</div >');
+            
+            $('#alertsModal').prepend(alert);
+            return;
+        }
+            //return;
+
+        $('#loadingSchedule').show();
+        $('#free_appointments').empty();
```
milana-todorovic: I would not switch the page on the modal before the loading is finished. Also, I'd disable the buttons except for Previous while loading.

HealthcareSystem/PatientWebApp/wwwroot/js/patient_examinations.js
```diff
@@ -386,4 +328,102 @@ function cancelExamination(id) {
 
 function FillOutTheSurvey(examinationId) {
     window.location.href = '/html/filling_out_the_survey.html?id=' + examinationId;
+};
+
+function step(id) {
+
```
milana-todorovic: Just curiosity, what is the advantage of this being one function instead of three? It behaves differently depending on the step anyway, and I don't see any common bits, why not just separate it?

milana-todorovic (commented):
It's okay, but I left some HCI suggestions in the code, and here are a couple more (listed in order of importance - the first two are really important, the rest are just aesthetic things):
- <strike>I don't really like where you're doing the validation. Why does it only alert me that I needed to select a date in the first step, when I click next in the third step where the doctor selection is? Why do I get the 'You must select a date' message if I have in fact selected a date, but haven't selected a doctor? Just don't let the Next execute in the first step until I put in a date and alert me, and don't let Next execute in the third step until I choose a Doctor and alert me.</strike> - fixed in 08ca7bf90e65a71d8842e5a6f24aeb0dded2614f.
- I'm not seeing an alert that scheduling was successful anywhere, it would be nice to have that. - reload added in 08ca7bf90e65a71d8842e5a6f24aeb0dded2614f.
- So this modal currently doesn't reset if I close it and reopen, and I'm honestly not sure if it should or not? I feel like it maybe should cause it's like, I already cancelled and the parameters I didn't want are still there (plus, the loading in appointments might easily become out of date). Seems achievable too, just put an on click handler on the Close button along with the data-dismiss. 
- I would probably reorganize this page to have the Schedule an appointment card above the instead of on the left, and center everything. It's too small now, putting it on the side doesn't really make sense.
- <strike>I kind of hate that the Previous button is yellow, it feels like it clashes with all the other colours. Maybe use btn-light instead, that should be a nice light gray? Previous is kind of a thing that should be inobtrusive anyway.</strike> - fixed in 08ca7bf90e65a71d8842e5a6f24aeb0dded2614f (thanks, I know this was the pettiest complaint ever but it was just burning my eyes)

HealthcareSystem/PatientWebApp/wwwroot/js/patient_examinations.js
draganacarapic: I have changed most things at your suggestion except the following which are not clear to me. I think alert that scheduling was successful is located on 114-116 lines on file patient_examination.js.  
"I would not switch the page on the modal before the loading is finished." Also this part is not the clearest to me. What should I do?

milana-todorovic: Oh, so right now, once everything is chosen, you switch to that last step right away, and an empty combo box is shown (along with that loading alert). I would maybe just stay on the previous step, show the loading alert, disable the Next button so the user knows it's already been clicked, and then switch to the step with the combo box with appointments once the AJAX is done and appointments are shown. 
Although I'm fine if you choose to skip with modification, it's more of a personal preference. I personally think it would be nicer but I can't really objectively back it up.

HealthcareSystem/PatientWebApp/wwwroot/js/patient_examinations.js
```diff
@@ -83,6 +83,9 @@ $(document).ready(function () {
         }
     });
 
+    $('#btn_close').click(function () {
+        location.reload();
+    });
```
milana-todorovic: Reloads are always a bit eh to me but it's alright.

milana-todorovic (dismissed):
The validation looks much better now, we could consider the rest of the suggestions if there's time (maybe if we do a bit of a general HCI cleanup before the demo?), but they're a lot less relevant.
Also, another not that relevant but maybe nice suggestion, maybe clicking on Next should also clear the alerts (since any validation errors have been fixed if Next went through)?

anciz (dismissed):

anciz (approved):

sladjana-savkovic (approved):

PR #212, opened 2021-01-18
refactor:modify PatientWebApp E2E tests

sladjana-savkovic:


milana-todorovic (approved):
Seems alright (double check that you didn't miss any of the urls though, I didn't do a detailed check). I'll go through and make the urls configurable sometime in the following days, I assume that you want these urls left as the default when that is done?

vlaksi (approved):

PR #226, opened 2021-01-25
feat[g5.1, g5.2, g5.3]: Scheduling renovations connected with frontend; Finished search for neighboring rooms for room merging

VasilijeBursac:
- This branch contains only **code** on Graphical Editor **Frontend**, so it really won't affect anything on the server and backend
- On this branch, I have connected "Scheduling basic renovations", "Scheduling complex renovations", "Entry of new information after renovation"  frontend with backend
- On this branch, I have completed and improved logic for finding neighboring rooms for room merging
(Note: Code for finding neighboring rooms is not the cleanest, but I have not found a proper way to split it in multiple functions because it wouldn't improve the situation that much and functions would have a lot of parameters)


VasilijeBursac:
[(2) g5.1: Schedule a basic renovation](https://trello.com/c/MO60PUeI/61-2-g51-schedule-a-basic-renovation)

VasilijeBursac:
[(3) g5.2: Schedule a complex renovation](https://trello.com/c/egMI7Yih/62-3-g52-schedule-a-complex-renovation)

VasilijeBursac:
[(3) g5.3: Enter new information after scheduling a renovation](https://trello.com/c/MBN2A72C/63-3-g53-enter-new-information-after-scheduling-a-renovation)

kmilena98 (approved):
Works well.

milana-todorovic (dismissed):
Seems alright, and as far as I've checked there are no issues with the rest of the system.

HealthcareSystemFrontends/GraphicalEditor/Service/MapObjectServices.cs
```diff
@@ -71,5 +73,56 @@ public List<MapObject> GetNeighboringRoomsForRoom(MapObject room)
...
+
+        private bool CheckIfMapObjectsOverlap(MapObject firstMapObject, MapObject secondMapObject)
+        {
+           for(double coordinateX = Canvas.GetLeft(firstMapObject.Rectangle); coordinateX <= (Canvas.GetLeft(firstMapObject.Rectangle) + firstMapObject.Rectangle.Width); coordinateX += firstMapObject.Rectangle.Width/5)
+            {
+                if (CheckIfPointInsideMapObject(new Point(coordinateX, Canvas.GetTop(firstMapObject.Rectangle)) , secondMapObject))
+                {
+                    return true;
+                }
+            }
```
milana-todorovic, line 85: So this wouldn't really fix that much, but maybe get rid of the brackets? And doing <code>if (condition) return true</code>, just do <code>return condition</code>.

milana-todorovic (approved):
Seems alright. You're right, the code for finding neighbors could be cleaner, but it's not as bad as it could be. I added a small suggestion bellow for how to remove some redundant bits.

HealthcareSystemFrontends/GraphicalEditor/Service/MapObjectServices.cs
VasilijeBursac: Thank you for your suggestions! I will remove brackets, but in terms of `return condition`, I would do that, but I need to return `true` only if that condition is `true`, because I need all of the condition checks to run, and If at least one of them is `true`, I will return `true`. On the other side, the function needs to return `false` only if none of the checks return `true`.

PR #229, opened 2021-01-25
refactor-ia-tests-and-mailing

borislav-puzic:
I want to refactor tests to work with microservice, and add email sending in case of invalid ActionBenefitMessage


milana-todorovic (commented):
Do the tests work fine locally and match what you and your product owner agreed was an acceptable way to do integration tests for this project? If so, I would probably temporarily remove them from the pipeline and merge this branch. And then we can open another branch tomorrow to try to figure out what needs to be running in the pipeline for the tests to pass and to set that up.

milana-todorovic (dismissed):
Seems alright.

vlaksi (dismissed):

Pufke (dismissed):

milana-todorovic (approved):

Stefan-Kitanovic (approved):

PR #231, opened 2021-01-25
chore[d74]: backend pruning

anciz:
All classes which are not used are deleted.


milana-todorovic (commented):
I think you removed all classes from some namespace that is no longer used, so the namespace doesn't exist anymore, but the imports were left. You'll have to go through and remove all the unused package imports (finish deleting classes from Backend first though, then loop through Build - Ctrl R+G in files with errors until the build passes.

vlaksi (dismissed):

milana-todorovic:
[(1) Backend pruning [d74]](https://trello.com/c/fr6wyHjD/74-1-backend-pruning-d74)

milana-todorovic (commented):
Could you delete the interfaces of the Service classes you deleted as well? And I think some repositories can be removed too. 
Also, check on FeedbackService, SurveyService and such, I don't think they're used anywhere? Or at least they shouldn't be.

anciz:
> 
> 
> Could you delete the interfaces of the Service classes you deleted as well? And I think some repositories can be removed too.
> Also, check on FeedbackService, SurveyService and such, I don't think they're used anywhere? Or at least they shouldn't be.

I deleted interfaces and classes of the Service I deleted. Also, I removed some repositories and other services. Also, I removed some classes from Model.

milana-todorovic:
Pretty much all files in the solution have been changes because I ran Visual Studio's Code Cleanup to remove unnecessary usings.

vlaksi (approved):

Pufke (approved):

PR #232, opened 2021-01-26
feat[g5.4, g5.5]: Room Schedule and Canceling Scheduled Actions

VasilijeBursac:
- This branch contains "Room Schedule" for selected room and "Canceling Scheduled Actions"
- Also, you can see more details about selected scheduled action in a separate dialog

- Dialog for more details about Equipment Relocation doesn't have any fields because we have to add more fields to the database table so we can return the complete object with all necessary Info for display in dialog.


VasilijeBursac:
[(3) g5.4: Show schedule for each room [2]](https://trello.com/c/pbzVn3Rk/64-3-g54-show-schedule-for-each-room-2)

VasilijeBursac:
[(2) g5.5: Cancel scheduled actions [2]](https://trello.com/c/BSVF1Qrt/65-2-g55-cancel-scheduled-actions-2)

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
```diff
@@ -100,6 +103,25 @@ public ActionResult DeleteEquipmentTransfer(int id)
             return Ok();
         }
 
+        /*
```
kmilena98, line 106: Why is this code commented? If it will not be used in the future, l suggest deleting it, since the last sprint is done.

kmilena98 (approved):

HealthcareSystem/GraphicalEditorServer/Controllers/EquipmentController.cs
VasilijeBursac: I will uncomment this once Marko adds all required fields to the Equipment Relocation table and creates DTO.

Dialog for more details about Equipment Relocation doesn't have any fields because we have to add more fields to the database table so we can return the complete object with all necessary Info for display in the dialog.

milana-todorovic (approved):
Seems alright.

